PR #3, opened 2022-05-17
Add GWI service

lkslts64:


main.go
```diff
@@ -0,0 +1,45 @@
...
+	// Run our server in a goroutine so that it doesn't block.
+	logger.Printf("Starting GWI service on port %s", *port)
+	go func() {
+		if err := svc.ListenAndServe(); err != nil {
+			logger.Println(err)
+		}
+	}()
+	c := make(chan os.Signal, 1)
+	// We'll accept graceful shutdowns when quit via SIGINT (Ctrl+C)
+	signal.Notify(c, os.Interrupt)
```
lmacko: good to signals handled 👍

service/handlers.go
```diff
@@ -0,0 +1,437 @@
...
+		if err != nil {
+			s.log.Printf("Error updating asset: %v", err)
+			http.Error(w, "Asset ID not found", http.StatusNotFound)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+	}
+}
+
+func (s *Service) handleDeleteAsset() http.HandlerFunc {
```
lmacko: why do you return `http.HandlerFunc` instead of implementing handler in `handleDeleteAsset` directly?

service/middleware.go
```diff
@@ -0,0 +1,86 @@
...
+				return
+			}
+			if !tkn.Valid {
+				w.WriteHeader(http.StatusUnauthorized)
+				return
+			}
+			// ==== User is now authenticated ====
+
+			// Check for ratelimit.
+			if !limiters[claims.Username].Allow() {
```
lmacko: Is this going to work for users other than the pre-created ones?
Scenario.:
```
1, POST /users
2, POST /login                 with the newly created user
3, GET /users/:id/favourites   authorized by newly created user's token 
```

service/storage.go
```diff
@@ -0,0 +1,290 @@
...
+// 	for _, u := range s.users {
+// 		if !f(u) {
+// 			return
+// 		}
+
+// 	}
+// }
+
+func (s *storage) getUsers() []*user {
+	s.usersMu.RLock()
```
lmacko: Is there any benefit of locking here if a pointer to "shared object" is returned?

service/types.go
```diff
@@ -0,0 +1,238 @@
...
+
+type gender string
+
+const (
+	male   gender = "male"
+	female gender = "female"
+)
+
+type user struct {
+	ID    uint   `json:"id"`
```
lmacko: Why does the unexported type `user` contains exported fields?

lmacko (commented):
Hi @lkslts64 thanks for submitting the challenge! Could you answer  questions in comments?

main.go
lkslts64: Thanks

service/handlers.go
lkslts64: I've borrowed this technique from [this](https://pace.dev/blog/2018/05/09/how-I-write-http-services-after-eight-years.html) blog post. Basically, you can use this if you want to do some setup/preparation before the handler code executes. It also gives the handler a closure environment to operate.  In this project, none of my handlers benefits from this pattern (only middleware code), but it has become for me the standard way of declaring handlers. So, instead of switching between different ways of declaring handlers, I prefer consistency and I use this pattern even if sometimes it is not needed.

service/middleware.go
lkslts64: No, it is not. An assumption I've made that I forgot to mention in README is that the pre-created users are unrelated to the ones that you can create through the API. There was no intention of this to be working for the API created users.

service/storage.go
lkslts64: No, and in fact, this is a terrible mistake. Race conditions could emerge since this code is not thread-safe but was intended to be. I should return `[]user` instead. Can I make a quick patch to the code or should I live it as is?

service/types.go
lkslts64: That's an easy one :). `json` package uses reflection to marshal the `user` struct and needs to have visibility of its fields. In Go, a package cannot view the unexported fields of types within a different package.

juanramoncg:
Hi @lkslts64 thanks for taking the time for doing the challenge. Instead of creating a PR against this repository, could you please do a Fork in your side and send us the link, granting us permissions to @lmacko for commenting there?
Maybe we can copy/paste opened comments in that Fork @lmacko and continue from there?

juanramoncg:
hey @lkslts64 thanks a lot for the time on the challenge. One thing I have noticed, could you instead of creating a PR against this repository create a PR against your fork and grant @lmacko access to it?
@lmacko do you think we can copy/paste opened comments to that PR and then continue from there?

lkslts64:
Hey @juanramoncg. Done and sorry for the inconvenience. I was told to "make a pull request" without any further details, and I was a little bit confused.

juanramoncg:
Thanks @lkslts64 no worries at all

PR #4, opened 2022-09-21
Implementation from Manos Ragiadakos

rm4n0s:
Hi GWI,

This is my implementation of the exercise.
It is missing some tests that I didn't have the time to implement.
Furthermore, it may contain some bugs in SQL.
Overall, I hope you enjoy my code.

Best regards,
Manos


sqldb/assets.go
```diff
@@ -0,0 +1,393 @@
...
+	}
+	assets := []domain.Asset{}
+	switch query.Type {
+	case domain.InsightAssetType:
+		ins := []Insight{}
+		err := gormQuery.Limit(query.Limit).Find(&ins).Error
+		if err != nil {
+			return nil, err
+		}
+		assets = listRowsToAssets(ins)
```
lunemec, line 163: How would you change the listing part of code in case you have millions of entities?

No need for working code, just a list of changes necessary with some hints as to why.

sqldb/assets.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func (d *DB) listFavouriteAudiences(ctx context.Context, userID uint, onlyFav bool, query domain.QueryAssets) (*domain.ListedAssets, error) {
+	var aus []AudienceWithFavour
+	gormQuery := d.db.Model(Audience{}).Select("audiences.*, (favourite_audiences.user_id = ?) AS is_favourite", userID)
+	if onlyFav {
+		if query.IsDesc {
+			gormQuery = gormQuery.Joins("INNER JOIN favourite_audiences ON favourite_audiences.audience_id = audiences.id AND audiences.id < ? AND favourite_audiences.user_id = ?", query.LastID, userID).Order("audiences.id desc")
+		} else {
+			gormQuery = gormQuery.Joins("INNER JOIN favourite_audiences ON favourite_audiences.audience_id = audiences.id AND audiences.id > ? AND favourite_audiences.user_id = ?", query.LastID, userID).Order("audiences.id asc")
```
lunemec, line 259: These 2 parts of the condition differ only in the comparison operator `< >` - how would you approach abstracting the query?

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM golang:1.19-alpine AS build
+WORKDIR /app
+COPY go.mod go.sum ./
+RUN go mod download
+COPY *.go ./
+COPY . ./
+RUN go build -o /gwiserver
+
+
+FROM alpine
```
lunemec, line 10: Can you describe why this pattern with builder image is used?

config.json
```diff
@@ -0,0 +1,9 @@
+{
+    "sql_db":{
+
+    },
+    "http_api":{
+        "port": 8000,
+        "jwt_secret":"secret"
```
lunemec, line 7: How would you go about configuring the app secrets in production environment?

lunemec (commented):
Very nice work 👌 
- auto generated swagger docs 👍 
- tested very well, and uses testify 👍 
- how-to run/test/... 👍 

I left few questions there

sqldb/assets.go
rm4n0s: First, I would remove the listRowsToAssets by removing the domain.Asset wrapper and refactoring the code to use the domain.AssetType for recognizing the type of the asset. 
The reason is that listRowsToAssets have O(n) debt and finding the asset type through switch v := asset.Data.(type) is slower.
Moreover, I would replace the Find method with FindInBatches and the Create method with CreateInBatches, as it is recommended from their "Performance" section of their documentation.

sqldb/assets.go
rm4n0s: When I started writing the SQL code, I was wondering if they were going to become multiline so I can put them in .tpl files and use the template package to add the values.
But I endured the long strings without thinking too much how to improve them because of the personal time restriction at the time of the development.
However, for this specific code I would rewrite it to something like this
```go
ineq := ">"
order := "asc"
if isDesc {
   ineq = "<"
   order = "desc"
}

gormQuery = gormQuery.Joins(fmt.Sprintf("INNER JOIN favourite_audiences ON favourite_audiences.audience_id = audiences.id AND audiences.id %s ? AND favourite_audiences.user_id = ?",ineq ), query.LastID, userID).Order("audiences.id "+order)
```

Dockerfile
rm4n0s: Yes, this pattern creates two images and moves the binary from the first image to the second image.
The reason behind this pattern is to build the code first in an image that contains the Go compiler, by moving the code and dependencies to that image. 
And after the compilation, we move the binary from the first image to the second image, which we want to be smaller and safer for production usage.

config.json
rm4n0s: Actually, I forgot to delete the config.json. 
I don't know what I was thinking when I created the config.json because it is a bad practice.

All the configurations are in .env and .env.docker (to be used from docker-compose or from independent docker image).
The reason that I prefer .env for production is that it moves the configurations to the environment of the OS for the executable only, making it very hard for the hacker to obtain them.
So in production, we can delete the secret from .env and add it in another way to the OS environment of the executable which is more secure, for example Kubernetes "Secrets" are similar to ConfigMaps but more secure.

rm4n0s:
Thank you for the exercise and the questions.
I am grateful for your patience, looking at the code and waiting for my answers (sorry for the delay, I took a nap).
Moreover, I learned a lot about my boundaries of my knowledge and skills with this exercise, which helped me a lot to grow as a developer.
In conclusion, I enjoyed every bit of the experience and I hope to be part of the team to keep the experience alive.

PR #5, opened 2022-10-24
Implementation

juanramoncg:


config/config.go
```diff
@@ -0,0 +1,62 @@
...
+func NewConfig() *Config {
+	return &Config{
+		MySQL: NewMySQLConfig(),
+		JWT:   NewJWTConfig(),
+	}
+}
+
+func NewMySQLConfig() *MySQLConfig {
+	u := os.Getenv("MYSQL_USER")
+	pw := os.Getenv("MYSQL_PASSWORD")
```
jhatzics, line 35: Is there any other way to protect the database password during the ci/cd process?

http/user/handlers.go
```diff
@@ -0,0 +1,191 @@
...
+		return err
+	}
+	if _, found := assets.AssetTypes[s.AssetType]; !found {
+		return fmt.Errorf("invalid asset type: %v ", s.AssetType)
+	}
+	return nil
+}
+
+// GetDashboard
+func (h *Handler) GetDashboard(w http.ResponseWriter, r *http.Request) {
```
jhatzics, line 50: How you could unit test the logic of this function? Are the any limitation with the current implementation?

cmd/platform-go-challenge/main.go
```diff
@@ -0,0 +1,158 @@
...
+			log.Print(err.Error())
+			return
+		}
+	}
+
+	_ = http.ListenAndServe(":5000", r)
+}
+
+func createRouter(services services, cfg config.Config) *mux.Router {
+	r := mux.NewRouter()
```
jhatzics, line 54: Why you did not choose a web framework? Have you tested any of the most popular options?

http/user/handlers.go
```diff
@@ -0,0 +1,191 @@
...
+	w.Header().Set("Content-type", "application/json")
+
+	userID, err := strconv.Atoi(mux.Vars(r)["user_id"])
```
jhatzics, line 53: Would be possible to send arbitrary userIds from the client and the app respond with the corresponding user data?

http/user/handlers.go
```diff
@@ -0,0 +1,191 @@
...
+
+func (h *Handler) RemoveFromDashboard(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-type", "application/json")
+	userID, err := strconv.Atoi(mux.Vars(r)["user_id"])
+	if err != nil {
+		http.Error(w, "wrong user id", http.StatusBadRequest)
+		return
+	}
+	action := starAction{}
+	err = action.DecodeAndValidate(r)
```
jhatzics, line 109: Would be possible to avoid this step from each handler? If yes, how would you approach it?

http/user/handlers.go
```diff
@@ -0,0 +1,191 @@
...
+	err = h.service.AddToDashboard(ctx, uint32(userID), action.AssetID, action.AssetType, action.Description)
+	if err != nil {
+		log.Print(err.Error())
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(http.StatusNoContent)
+}
+
+func (h *Handler) RemoveFromDashboard(w http.ResponseWriter, r *http.Request) {
```
jhatzics, line 101: Why are you using a pointer receiver here?

internal/repositories/sql/dashboards.go
```diff
@@ -0,0 +1,220 @@
...
+func NewDashboardsRepository(dbClient BasicConnectionWithTransactions) *Dashboards {
+	return &Dashboards{dbClient: dbClient}
+}
+
+// GetUserDashboardID returns id of a user's dashboard.
+func (d *Dashboards) GetUserDashboardID(ctx context.Context, userID uint32) (uint32, error) {
+	dashboardID := uint32(0)
+	err := sq.Select("d.id").
+		From(dashboardsTable + " AS d").
+		Where(sq.Eq{"d.user_id": userID}).
```
jhatzics, line 30: Could that lead to SQL Injection?

internal/app/dashboards/model.go
```diff
@@ -0,0 +1,39 @@
...
+type StarredCharts []StarredChart
+
+type StarredInsight struct {
+	assets.Insight
+	Description string `json:"description"`
+}
+
+type StarredInsights []StarredInsight
+
+type StarredAudience struct {
```
jhatzics, line 27: Do you see any issues using a model for both DB and HTTP interfaces? How you could separate those two entities if it was necessary? What would change on both HTTP and Repository ends?

Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM golang:1.18-rc-alpine3.14
+
+RUN mkdir /platform-go-challenge
+COPY . /platform-go-challenge
+WORKDIR /platform-go-challenge
+RUN go build cmd/platform-go-challenge/main.go
```
jhatzics, line 6: Would be possible to do something here to improve the performance of the final app image?

jhatzics:
Thank you very much Apostole for the effort you put into this assignment. I could say it is a nice work and quite Go idiomatic. It was nice reviewing it. Apart from the comments, I would like you also to answer the following questions,

* What extra implementation would you put in if you had more time (e.g. from a security perspective, or anything else)? 
* If you imagine this service to serve millions of users, what would you change to make it more performant?

Thank you once again! Nice work mate!

config/config.go
apopsallas: I 'm a bit weak in DevOps like processes but: 
i. I would not put this information on the repository.
ii. I would use different credentials for different stages in development.
iii. In my understanding even if the credentials where just in a .env file in a server somewhere there is still a chance to be "seen" through a tool like Sentry in the log stack trace. I believe there are solutions provided by third party companies to keep this kind of credentials safe.

http/user/handlers.go
apopsallas: I usually run end to end tests for handlers but I have used the `net/http/httptest` package in the past and I could mock the services passed in the Handler struct. 
Unit testing really helps with understanding if the implementation is well written or not so if I had done unit testing for this I would be sure if there were any limitations. I can't see it right now.

cmd/platform-go-challenge/main.go
apopsallas: I figured it would be an "overkill" for a simple CRUD functionality.
Plus I felt like it would be easier for me since I haven't used a GO framework recently. I have only tried GIN 2-3 years ago.

http/user/handlers.go
apopsallas: Yes. This endpoint is not protected whatsoever. I only added some kind of protection to "write" endpoints but I guess I should have done the same with this one.

http/user/handlers.go
apopsallas: I would create a custom unmashal method for the request body and would probably add an extra middleware to not have to do this kind of checks inside the handler.

http/user/handlers.go
apopsallas: Actually there is no need for a pointer receiver there since I don't want to "mutate" the struct or anything of that sort. It's a mistake.

internal/repositories/sql/dashboards.go
apopsallas: I purposely used this package since it's easier to build a safe sql query. It keeps all arguments in a slice and uses placeholders in the generated query.

internal/app/dashboards/model.go
apopsallas: Yes. Since we may have some struct fields that we don't want to return or the actual structure of the model may be different in many ways we should have different models sometimes even per call or per end user. 
I would most probably create view models with custom marshal methods and constructors that used the DB models as parameters. The end result would have any kind of structure or fields we needed without having to use different implementations each time.

Dockerfile
apopsallas: The only thing I can think of at the moment is to have a more lightweight base image from the start which is why I picked this one.
I am weak in this kind of stuff and I would love to learn more about them and gain hands on experience.

apopsallas:
I thank you for you remarks. They were quite helpful and made me think a bit more about my implementation and my skills as a developer.
I also thank you for your kind words and the time you put in checking out my code!

> What extra implementation would you put in if you had more time (e.g. from a security perspective, or anything else)?

I believe I had plenty of time for this assignment I just had a late start. 
* Probably have more tests (unit and end to end) instead of just some integration tests.
* Definitely would add missing comments.
* A better business logic approach because the dashboard functionality felt a bit limiting the way I implemented it.
* I thought of adding some concurrency logic in the repository to fetch the different assets concurrently instead of one type at a time.
* Caching would be an extra functionality I was thinking of adding.
* Probably swagger and a better documentation.

> If you imagine this service to serve millions of users, what would you change to make it more performant?
* I would definitely add a cache mechanism.
* Maybe some concurrency logic in the repository as I wrote above.
* I would investigate in the benefits of a NoSQL DB for performance reasons. 
* Would ask for help from my team leader :P

PR #6, opened 2022-11-21
Gwi challenge implementation

ghal:
## Problem
Create a favorite asset service that contains information about the user's favorite assets.

## Solution
This PR consists of 2 commits:
1. The implementation of the challenge(https://github.com/GlobalWebIndex/platform-go-challenge/pull/6/commits/dbedf22c3788aec0d07d27b89f440837e4627c0a).
2. The vendor folder (in order to keep vendor historicity)(https://github.com/GlobalWebIndex/platform-go-challenge/pull/6/commits/464781eaf3b1c79e7d6d9b7b5ca53cd03255d95d).


infra/deploy/build/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM golang:1.18 as builder
+WORKDIR /home/platform-go-challenge
+COPY . ./
+
+RUN CGO_ENABLED=0 GOOS=linux go build -mod=vendor -a -installsuffix cgo -ldflags "-X main.version=$version" -o platform-go-challenge ./cmd/platform-go-challenge/main.go
```
lunemec: Why did you choose to use `-mod=vendor`? Also why do you have `CGO_ENABLED=0` and then `-installsuffix cgo` ?

internal/infra/repository/mongo/favouriteasset/cacherepository.go
```diff
@@ -0,0 +1,95 @@
...
+const CacheTTL = 30 * time.Second
+
+// CRepo is the favouriteasset repo with cache functionality.
+type CRepo struct {
+	repo        Repo
+	redisClient *redis.Client
+}
+
+// NewCacheRepository constructor.
+func NewCacheRepository(repo Repo, redisClient *redis.Client) (*CRepo, error) {
```
lunemec, line 24: I think there is a bug in the cache repository (no need to fix it), but how would you write a test to discover it and fix it?

Similarly, if you wanted to upgrade to new mongo DB version, what tests would you add in order to be sure there isn't any issue with the upgrade?

README.md
```diff
@@ -1,31 +1,96 @@
...
+- [ ] Use builder pattern for creating objects.
+- [ ] Add metrics.
+- [x] Docker implementation.
+- [x] Postman collection.
+
+## Future Improvements
+
+- In the current implementation Assets & Favourite assets are coupled into the same db. In order to decouple these, a
+  new
+  service should be created that aggregates assets & favourite assets along with users using the BFF pattern.
```
teliaz, line 87: What would be the deciding factor to split a domain in to multiple microservices?

cmd/platform-go-challenge/main.go
```diff
@@ -0,0 +1,85 @@
...
+	err := godotenv.Load("../../.env")
+	if err != nil {
+		e.Logger.Warnf("no .env file exists: %v", err)
+	}
+
+	// Setup app config
+	cfg := config.New()
+
+	// set up Mongo.
+	mongoClient, err := setUpMongoDB(ctx, cfg)
```
teliaz, line 43: Are you missing a error check here?

doc/diagrams/Platform-Go-Challenge.drawio
```diff
@@ -0,0 +1 @@
...hkdId6RWO" name="Page-1">5VnbctowEP0aHtOxfCHkMYH0NkmbCdNp8tQR9sZWInupLC7u11fC8g0TSjoQk/aBGe/qgnXOnmNY95xhvPwg6DS6xgB4z7ZCwYKeM+rZNlEflZjSEBoJPWPMfhVJy2RnLIC0MVEicsmmzaSPSQK+bOSoELhoTntA3r6NsU85tLLfWSCjPDvwrCr/EVgYFV9ELDMS02KySaQRDXBRSzmXPWcoEGV+FS+HwDUyBS75uvfPjJY3JiCROy0I7764aQbffjyyTz/J/Wfk6QnxzM3JrDgxBAoAE6KQEYaYUH5ZZS8EzpIA9LaWiqo5V4hTlSQq+QhSZoY8OpOoUpGMuRmFJZN3evk7z0T3jWi0NHuvgqwW3IBgMUgQRS6RIrurB7WddFhttYqKvfJT66M+i6ZJpTgTPmyD0BQhFSHILfPcknOlBEB1CpGpdQI4lWzevA9qijQs51XEqgvD7Ut4zvedUz4z33TuSxQt8lWRTvXlLOb5BOdiDkIyJYgrOgF+gymTDBM1ZYJSYlybcM5ZqAekroI63TiTnCUwLBVZMaDXwnI7B23MzIK+UVaWh06htEUlVMekoppGi2V7B9lugXwLAUufBdnPFCwBCEfBsYiYhPGUrkptoYyyCeEkl9zVpExQ/ylcCfFrjq7Jp7nmiLcfiE+9JsbkrI1xfwPGg0Nh7LQwvsYkxDeN8Vodu07HGBO3y4cCeckjoXv7d3e0f69L+3fbzqRMGdr+32TxD4LZQ+2THTyc2K9p4mfHU/unR1/83o7FP+j0t491PJTa/wyl/S4p9Vp+NgYxZ+pgXRtaaWBHY2ikG0crK7kq3vtG7R68kvtvwZzsIzKn7d7UATN2l8z0Wx4zpH6kHMa6hWn7D8drG423ZjS21bXRlPQffSn/7WN2Fa23oPaoi8GOunC61MWgpYuj1IPrdq0H0m66aQdhSajrJaETruBZh00BIJvYUNNW8xUQUG/IFf22mAVBriVI2S+9rQF9iiyRq1N5Fz1vpPdS8im6F7qRIQU+wRC5bvSNEtQdjosHxvlaSqCkput3cran5p1L1vyrv4EvsoEv52B8tft3FV8BS/93wiyvKTCyo8AORpjdJow9qDhB2dNvdpQBvTm69sWWt95Y2dBUHOyHLBVWb7FWY7UXfc7lbw==</diagram></mxfile>
```
teliaz, line 1: Thanks for providing these diagrams!

internal/app/favouriteasset/service_test.go
```diff
@@ -0,0 +1,298 @@
...
+
+func TestNewService(t *testing.T) {
+	repoMock := &rfavouriteasset.RepoMock{}
+	s := favouriteasset.NewService(repoMock)
+
+	assert.IsType(t, new(favouriteasset.Service), &s)
+}
+
+func Test_assetService_GetFavouriteAssets(t *testing.T) {
+	favouriteAssetID := "favourite asset ID"
```
teliaz, line 22: Can you describe the AAA pattern on some of the tests?

internal/infra/http/router/favourite/router.go
```diff
@@ -0,0 +1,82 @@
...
+// NewRouter returns an HTTP component to serve all the routes for the user favourites.
+func NewRouter(asSvc favouriteasset.Service) *Router {
+	return &Router{
+		asSvc: asSvc,
+	}
+}
+
+// AppendRoutes adds favourite assets routes to router.
+func (r *Router) AppendRoutes(e *echo.Echo) {
+	e.GET("/users/:user_id/favourites", r.GetUserFavourites)
```
teliaz, line 24: No need to change anything. Would these routes be better if user was taken from token payload?

teliaz (commented):
Thank you for submitting the test! Can you answer the following alongside my comments? 

Thanks in advance and again thanks for the effort you put in this assignment!

infra/deploy/build/Dockerfile
ghal: Hello @lunemec, thanks for the feedback!

When the -mod=vendor flag is used, go bulds the package using the vendor directory. There are 2 other options but they ignore the vendor directory. As I see prior to go 1.14 the -mod=vendor is the default value so it's not needed.

The reason CGO_ENABLED=0 is that we do not use any c code in our repo. CGO_ENABLED env variable is enabled by default and it can have an overhead in resources. 

As I searched, `installsuffix ` workaround was useful in go versions < 1.10 where it was an issue with static files that were not being regenerated.  Now the issue is fixed and that flag can be removed since we are on v1.18

more info here: (https://github.com/golang/go/issues/9344#issuecomment-69944514 )
and here: (https://twitter.com/bquenin/status/1131304803160604672) :laughing:

README.md
ghal: Hello, @teliaz thanks for the feedback!

Firstly, performance and scalability issues. Secondly, the teams that may work in a big repo may face code conflicts during development, especially in refactorings, and this makes the development harder and less clean.

doc/diagrams/Platform-Go-Challenge.drawio
ghal: You are welcome :)

internal/app/favouriteasset/service_test.go
ghal: Yes, this is an example struct I'm using for all the table-driven tests:



	tests := map[string]struct {
                // definition part
		repoMock *rfavouriteasset.RepoMock
		ctx      context.Context
                 ...
	}{
		"Should get favourite assets": {
                  repoMock: ...
                  ...
		 // Mocking dependencies

		},
	}
	for name, tt := range tests {
		t.Run(name, func(t *testing.T) {

                         // Arrange
			aSvc := favouriteasset.NewService(tt.repoMock)

                         // Act
			res, err := aSvc.GetFavouriteAssets(tt.ctx, tt.userID)

                         // Assert
			assert.Equal(t, tt.expRes, res)
			assert.Equal(t, tt.expErr, err)
		})
	}

internal/infra/http/router/favourite/router.go
ghal: Of course, especially if the URLs are exposed. If this repo is part of a microservice that is hidden by a BFF, then another service is needed to authenticate the user and hit us with the id of the user in order to get the info needed.

internal/infra/repository/mongo/favouriteasset/cacherepository.go
ghal: Nice catch! 

First of all, I'm not handling the error case of the Redis FlushDBAsync call. Should create an error channel and fill it into the goroutine. Also, I could make the call run synchronously (without goroutines), but it will have a performance overhead since the program will be blocked during the call.

Secondly, I propagate the repo error directly to the above level, and with this, I will not know if the error is coming from the repo level or the caching level during testing.

cmd/platform-go-challenge/main.go
ghal: Yes, fixed it. That's where tests are useful!  :) Unfortunately due to lack of time I missed some tests.

ghal:
> Thank you for submitting the test! Can you answer the following alongside my comments?
> 
> * Are you following a specific design pattern on you submission?
> * How do you value code comments and when do you think that they are necessary?
> 
> Thanks in advance and again thanks for the effort you put in this assignment!

Thanks for your feedback, I have a lot of suggestions to add but unfortunately, due to lack of time I could do better. I've suggested some at the end of the Readme.md. 

I've used the following patterns:
- Service-Repo for structuring the business & storage logic.
- DI for passing the dependencies to constructors.
- Adapter for converting the dynamic mongo data to structs.
- Decorator that adds a layer between service & repo for caching the repo results.

Regarding the comments question, I think that is up to the team to choose the amount of commenting that a repo can have. Ideally, the code should be self-explanatory with proper namings, good documentation, and diagrams.  Also, design patterns can help developers communicate the overview of the implementation.

PR #7, opened 2023-04-25
Go platform implementation

pdimiropoulou:


utils/auth.go
```diff
@@ -0,0 +1,64 @@
...
+			} else {
+				errorm.Message = "Unauthorized Access"
+				SendError(w, http.StatusUnauthorized, errorm)
+				return
+			}
+		} else {
+			errorm.Message = "Unauthorized Access"
+			SendError(w, http.StatusUnauthorized, errorm)
+			return
+		}
```
lkslts64, line 62: How would you refactor this code so as to remove both `else` blocks?

utils/database.go
```diff
@@ -0,0 +1,28 @@
...
+	"fmt"
+	"log"
+	"os"
+
+	"gorm.io/driver/postgres"
+	"gorm.io/gorm"
+)
+
+var DB *gorm.DB
+var err error
```
lkslts64, line 13: Defining exported global variables can be really handy in some situations. However, it comes with many downsides. Can you mention some of them? How would you approach this if you weren't allowed to use global variables?

repository/asset_psql.go
```diff
@@ -0,0 +1,73 @@
+package repository
+
+import (
+	"platform-go-challenge/models"
+)
+
+type AssetRepository struct{}
```
lkslts64, line 7: What is the purpose of this empty struct type? Had you added some fields in this struct, what they would have been?

README.md
```diff
@@ -2,30 +2,228 @@
 
 ## Introduction
 
-This challenge is designed to give you the opportunity to demonstrate your abilities as a software engineer and specifically your knowledge of the Go language.
+This solution provides user endpoints with basic authedication functionality along with a CRUD API for user assets. For this challenge a postgres database with sample data has been created in https://api.elephantsql.com/.
```
lkslts64, line 5: Nice one!

models/audience.go
```diff
@@ -0,0 +1,11 @@
+package models
+
+type Audience struct {
+	Asset
```
lkslts64, line 4: Nice use of type embedding!

controllers/asset.go
```diff
@@ -0,0 +1,67 @@
...
+type Controller struct{}
+
+var assets []models.Asset
+var assetresp []models.AssetResponse
+
+func (c Controller) GetUserAssets(db *gorm.DB) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var error models.Error
+		limit_param := r.URL.Query()["limit"]
+		offset_param := r.URL.Query()["offset"]
```
lkslts64, line 25: What will happen if either `limit` or `offset` are not provided as query parameters?

controllers/audience.go
```diff
@@ -0,0 +1,68 @@
...
+type AudienceController struct{}
+
+func (a AudienceController) UpdateAudience(db *gorm.DB) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var audience models.Audience
+		var error models.Error
+
+		json.NewDecoder(r.Body).Decode(&audience)
+
+		if reflect.ValueOf(audience.ID).IsZero() || reflect.ValueOf(audience.UserId).IsZero() {
```
lkslts64, line 25: Is the use of reflection necessary here? Is there any alternative way to perform those validations?

lkslts64 (commented):
Hey @pdimiropoulou thanks for submitting the challenge! Could you please answer the questions in the comments?

pdimiropoulou:
> Hey @pdimiropoulou thanks for submitting the challenge! Could you please answer the questions in the comments?

Hey @lkslts64, thank you for your time on reviewing my PR! I will provide the answers shortly :)

utils/auth.go
pdimiropoulou: Valid point, I could have used `return` in the if/else statements in order the code to be more readable. It should look like the following 

                 if r.Header["Authorization"] != nil {

			token, err := jwt.Parse(r.Header["Authorization"][0], func(token *jwt.Token) (interface{}, error) {
				if _, ok := token.Method.(*jwt.SigningMethodHMAC); !ok {
					return nil, fmt.Errorf("Error on token")
				}

				return []byte(encryptionKey + id), nil
			})

			if err != nil {
				errorm.Message = "Unauthorized Access"
				SendError(w, http.StatusUnauthorized, errorm)
				return
			}

			if token.Valid {
				endpoint.ServeHTTP(w, r)
				return
			} 	
				errorm.Message = "Unauthorized Access"
				SendError(w, http.StatusUnauthorized, errorm)
				return
			
		} 
			errorm.Message = "Unauthorized Access"
			SendError(w, http.StatusUnauthorized, errorm)
			return

utils/database.go
pdimiropoulou: Some of the downsides using global variables are:

- Global variables have no access-control. We can access and change these variables in any part of the code, 
 thus it is difficult to debug.
- Global variables are not in harmony with Open-Close design SOLID principle 

Alternatively, I could have created struct and a repository interface in that will hold the the database dependency.

    type Database struct{
       Db *gorm.DB
    }
    
    func DatabaseInterface(db *gorm.DB) *Database {
       return &Database{Db: db}
    }

and use them in corresponding functions

repository/asset_psql.go
pdimiropoulou: My initial thought was to have a more flexible `AssetRepository` struct since the schema is not defined. More over, an empty struct has zero size and all pointers to it always point to the same special place in memory.

controllers/asset.go
pdimiropoulou: As it is now if one of them is missing, all user assets will be returned without any pagination. This should be changed in order to take into consideration either the limit or the offset if provided

controllers/audience.go
pdimiropoulou: It is not necessary at all, I could have used any other go validation package (ex: `validator`) or even write some validation rules/functions on my own by just checking the fields.

models/asset.go
```diff
@@ -0,0 +1,7 @@
+package models
+
+type Asset struct {
```
dzacharakis, line 3: In general, the goal of database design is to generate a set of tables allowing us to store information without unnecessary redundancy, yet also to retrieve information easily. Do you think `Asset` satisfies this goal? Why?

models/insight.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+type Insight struct {
+	Asset
```
dzacharakis, line 4: The relational schema describes the data and their relationships. As I can understand you have a superclass-subclass relationship where `charts`, `audiences`, `insights` are a specialization of a higher-level entity set `assets`. Is this relationship depicted in the schema?

models/asset.go
```diff
@@ -0,0 +1,7 @@
...
+	ID        uint `gorm:"primary_key;"`
+	UserId    int  `gorm:"not null;"`
```
dzacharakis, line 5: Consistency with data is really important, the data values stored in the database must satisfy certain constraints. For example, the `user` listed for each `asset` must be one that actually exists. Do you think `assets` satisfies this? Why?

models/insight.go
```diff
@@ -0,0 +1,6 @@
+package models
+
+type Insight struct {
```
dzacharakis, line 3: After connecting to your database:
```
baprmzsc=> SELECT * FROM assets;
 id | user_id | favourite 
----+---------+-----------
(0 rows)

baprmzsc=> SELECT * FROM insights;
 id | user_id | favourite |                                text                                 
----+---------+-----------+---------------------------------------------------------------------
  4 |       1 | t         | 40% of millenials spend more than 3hours on social media daily
  5 |       1 | t         | 50% of females in age 30-35 spend more than 8hours on running weekl
  6 |       1 | f         | 60% of females in age 50-60 spend more than 3 on internet daily
(3 rows)
```
Can you please elaborate on how table `assets` is utilized by your application?

dzacharakis (commented):
Thank you for your submission @pdimiropoulou! We really appreciate the time you took for this. Can you please also answer to me the following questions?

Thanks

models/asset.go
pdimiropoulou: Asset table as it is now doesn't really store any information. You can find details about this in my next answers, please let me know if something is not covered.

models/insight.go
pdimiropoulou: My initial though was to create `Assets` table as a parent table for `charts`, `audiences`, `insights` in order to access one table for favorites' assets. While implementing though, I realize that this wouldn't work better than having flag on each asset table (chart, insight, audience). I think I should delete this table since it is not really used anywhere and also it's not part of the migration

    func InitializeData() {
	    utils.ConnectPostgresDB()
    
	    utils.DB.AutoMigrate(&models.User{}, &models.Chart{}, &models.Insight{}, &models.Audience{})
    
    }

models/asset.go
pdimiropoulou: I should have created a foreign key to `users` in order to ensure data integrity. Just a note though, for sure the FK is needed but the validation if a user exists or not most probably will take place on a middleware.

models/insight.go
pdimiropoulou: Initially it was part of the DB migration, while implementing the REST API though this table wasn't really in use that's why it is empty. It should have been deleted from the DB

pdimiropoulou:
> Thank you for your submission @pdimiropoulou! We really appreciate the time you took for this. Can you please also answer to me the following questions?
> 
> Thanks

Hey @dzacharakis , thank you very much for your time on reviewing my PR. Please find my answers on comments, let me know if something is missing :)

utils/auth.go
lkslts64: Exactly! That is what I was looking for 🚀

utils/database.go
lkslts64: Great points made!

controllers/asset.go
lkslts64: Valid point!

controllers/audience.go
lkslts64: Great, that makes sense.

lkslts64:
@pdimiropoulou Thanks for answering my questions!

dzacharakis:
> Please find my answers on comments, let me know if something is missing :)

I'm fine, thanks again @pdimiropoulou for your effort!

models/asset.go
dzacharakis: With this comment, I wanted to highlight the lack of normalization in this database design. 

Treating `favourite` as an entity permits `users` to have several `assets` (including zero) associated with them.
Treating `favourite` as an attribute implies that `assets` have precisely one favourite each (by one `user`). 

Let's say we have another two users (user_id: 2, user_id: 3) that want to have the same insight as a favourite.
```
baprmzsc=> SELECT * FROM insights;
 id | user_id | favourite |                                text                                 
----+---------+-----------+---------------------------------------------------------------------
  4 |       1 | t         | 40% of millenials spend more than 3hours on social media daily
  8 |       2 | t         | 40% of millenials spend more than 3hours on social media daily
  9 |       3 | t         | 40% of millenials spend more than 3hours on social media daily
(5 rows)
```

The actual entity's info, the text, will be stored as many times as the number of users that will want this as favourite (redundancy).

pdimiropoulou: I see your point now, in that case we will need a reference table (let's call it `assets` ) that will hold the relationship between `assets` and `users`. By doing that, we can remove `favourite` property from the corresponding tables and add it to `assets` table. When a user flag/unflag an asset as favourite, `assets` table will be updated.

PR #8, opened 2023-05-01
GWI - Platform Go Implementation

vpagal:


repository/assetReporitory.go
```diff
@@ -0,0 +1,57 @@
...
+
+	qry := "INSERT INTO tAsset (description, type, data) VALUES ($1, $2, $3)"
+
+	result, err := db.Exec(qry, asset.Description, asset.Type, asset.Data)
+
+	if err != nil {
+		return err
+	}
+
+	asset_id, err1 := result.LastInsertId()
```
lkslts64, line 48: Assuming we can have many concurrent clients trying to insert assets, is `LastInsertId()` guaranteed to return the previously inserted Asset ID? If not, can you provide an alternative method for achieving this?

integrations/database.go
```diff
@@ -0,0 +1,37 @@
...
+	_ "github.com/lib/pq"
+)
+
+var host = os.Getenv("DB_HOST")
+var port = os.Getenv("DB_PORT")
+var user = os.Getenv("DB_USER")
+var password = os.Getenv("DB_PASSWORD")
+var dbname = os.Getenv("DB_NAME")
+
+var DB *sql.DB
```
lkslts64, line 17: Defining exported global variables can be really handy in some situations. However, it comes with many downsides. Can you mention some of them? How would you approach this if you weren't allowed to use global variables?

Dockerfile
```diff
@@ -0,0 +1,26 @@
+# syntax=docker/dockerfile:1
```
lkslts64, line 1: It's always nice to have a Dockerfile. Good job!

server/assetHandlers.go
```diff
@@ -0,0 +1,73 @@
...
+	"platform-go-challenge/repository"
+
+	"github.com/gorilla/mux"
+)
+
+func GetFavoriteForUserByIdHandler(w http.ResponseWriter, r *http.Request) {
+	params := mux.Vars(r)
+	userId := params["user_id"]
+
+	fmt.Println("GetFavoriteForUserById endpoint called for user with id : " + userId)
```
lkslts64, line 18: Instead of using `fmt` for logging information, can you think of another package intended for this purpose? What are the advantages of using that over `fmt`?

models/models.go
```diff
@@ -0,0 +1,38 @@
+package models
+
+type User struct {
+	Id       string `json:"user_id"`  // user id is actually the keycloak user id
+	Username string `json:"username"` // this should be the keycloak username
+	Password string `json:"password"`
+	Email    string `json:"email"`
+}
+
+type Asset[T any] struct {
```
lkslts64, line 10: Given the fact that you always instantiate `Asset` with the empty interface type argument (i.e `any`), is the use of generics justified here?

load_initial_data.sql
```diff
@@ -0,0 +1,41 @@
...
+    CONSTRAINT cUser_u1_user_id_username
+        UNIQUE (user_id)
+);
+
+DROP TABLE IF EXISTS tAsset;
+CREATE TABLE tAsset (
+    asset_id SERIAL PRIMARY KEY,
+    description VARCHAR(256),
+    type VARCHAR(32) NOT NULL,
+    data JSON NOT NULL,
```
dzacharakis, line 15: Nice, putting JSON type surely helps a lot, and query access for JSON documents in the context of relational DBMSs is great, but I am not sure if this is introduced in the SQL standard, meaning that it may be supported by many database systems (such as PostgreSQL) but not all. What could have been done to keep this idea but utilizing traditional SQL?

load_initial_data.sql
```diff
@@ -0,0 +1,41 @@
+DROP TABLE IF EXISTS tUser;
+CREATE TABLE tUser (
+    user_id VARCHAR(256) NOT NULL,
```
dzacharakis, line 3: Nice having the ID type as a non-guessable value. Can you think of another data type that could be more appropriate for this purpose?

repository/userRepository.go
```diff
@@ -0,0 +1,76 @@
...
+	qry := "SELECT user_id, username, email FROM tUser"
+
+	users, err := db.Query(qry)
+
+	if err != nil {
+		// handle this error
+		panic(err)
+	}
+
+	usersFound := make([]models.User, 0)
```
dzacharakis, line 22: If instead of this declaration, had used the following one, could have this made any difference in the slice?

```go
	var usersFound []models.User
```

repository/userRepository.go
```diff
@@ -0,0 +1,76 @@
...
+
+	qry := "SELECT user_id, username, email FROM tUser where user_id = $1"
+
+	user := db.QueryRow(qry, id)
+
+	userFound := models.User{}
+
+	var user_id string
+	var username string
+	var email string
```
dzacharakis, line 49: It is possible to declare a set of variables in a single declaration, this reduces these statements into one, but there is more room for improvement. How?

dzacharakis (commented):
Thank you for your submission  @vpagal! We really appreciate the time you took for this. Can you please answer the following questions? 

Thanks

lkslts64 (commented):
Thanks for submitting the challenge @vpagal! Could you please answer the questions in the comments?

load_initial_data.sql
vpagal: You are correct, this should be a UUID as the key provided by keycloak is a UUID actually.

load_initial_data.sql
vpagal: Using standar SQL we could create one table per asset, so one for Chart, one for Insight and one for Audience, those tables could look like :
`CREATE TABLE tChat (
    chart_id SERIAL PRIMARY KEY,
    description VARCHAR(256),
    x_axis_title VARCHAR(64) NOT NULL,
    y_axis_title VARCHAR(64) NOT NULL,
    data TEXT NOT NULL -- Not sure what exact type of data those should be
);`

and the connections with user caould happen again via an intermediate table like :
`CREATE TABLE tFavorite (
    user_id UUID NOT NULL,
    asset_type VARCHAR(32) NOT NULL,
    asset_id BIGINT NOT NULL,
    CONSTRAINT  PRIMARY KEY(user_id, asset_type, asset_id),
    CONSTRAINT cFavorite_f1 FOREIGN KEY (user_id)
        REFERENCES tUser (user_id),
    CONSTRAINT cFavorite_c1
        CHECK (asset_type in ('CHART', 'INSIGHT', 'AUDIENCE'))
);`

Ofc we will have to adjust accordingly or model structs in order to fit with this DB structure.

repository/userRepository.go
vpagal: I think both declarations will actually will produce a zero length slice, however this declaration :
`var usersFound []models.User`
will point to nothing. So the difference will be on serialization (if needed) where the declaration :
`usersFound := make([]models.User, 0)`
will produce an empty list : []
while the 
`var usersFound []models.User`
will return null.

repository/userRepository.go
vpagal: That's true, I believe I could use the var declaration block e.g. :
`var (user_id, username, email string)`

repository/assetReporitory.go
vpagal: If I got the point correctly I believe we could use the RETURNING option in our insert query in order to return the id of the current transaction.

integrations/database.go
vpagal: Of course it's a very good practise :) and tbh I wouldn't leave this as is if I had the familiarity with Go and a bit more time.
Using Globals which are shared between different packages/files make the code difficult to maintain and it's possible to lead to several defects.
Each time you have to make a change related to this variable you have to check every file using it making sure that won't affect/break it.
I would say that proper way of handling this would be changing the scope of the variable to local for this file and using a getter function in order to make it accessible outside of the package.

server/assetHandlers.go
vpagal: Ofc, I should use log (I think used "github.com/sirupsen/logrus" in same of my files :) ).
log packages provide us with many options on how to format or store our logs in one place so will be easier to access and utilize them when needed.

models/models.go
vpagal: Of course you are right :) I think you mean something like this :
`type Asset[T Chart | Insight | Audience] struct {`

repository/userRepository.go
dzacharakis: Nice, based on the zero-value mechanism the zero value of a slice type is nil.

repository/userRepository.go
dzacharakis: > I believe I could use the var declaration block

What I had in mind was this, but also yours reduces these statements into one.
```go
var user_id, username, email string
```
However, I wasn't talking about this. You see those variables could have been omitted.
```go
	userFound := models.User{}

	switch err := user.Scan(&userFound.Id, &userFound.Username, &userFound.Email); err {
	case sql.ErrNoRows:
		fmt.Println("No rows were returned!")
	case nil:
		fmt.Println(userFound.Id, userFound.Username, userFound.Email)
	default:
		panic(err)
	}
```

lkslts64:
Thanks again @vpagal for answering my questions!

server/userHandler.go
```diff
@@ -0,0 +1,87 @@
...
+		w.WriteHeader(http.StatusNotFound)
+		fmt.Fprintf(w, "User with that Id was not found")
+		return
+	}
+
+	jsonDoc, err := json.Marshal(&models.User{
+		Id:       userId,
+		Username: "Bill",
+		Email:    "pagalosb@gmail.com",
+	})
```
dzacharakis, line 76: What does this actually do? Why do you have hard-coded value? 

If I make a request:
`curl --location 'http://localhost:8080/users/0'`

The response will be:
```go
{
    "user_id": "0",
    "username": "Bill",
    "password": "",
    "email": "pagalosb@example.com"
}
```
But you initialized your database using: 
```sql
INSERT INTO tUser (user_id, username, email) VALUES ('afefr2345sdfs', 'Bill', 'bill@gmail.com');
INSERT INTO tUser (user_id, username, email) VALUES ('gdfg90908', 'John', 'john@gmail.com');
INSERT INTO tUser (user_id, username, email) VALUES ('842wf887ghsfsdf', 'Nick', 'nick@gmail.com');
```

load_initial_data.sql
```diff
@@ -0,0 +1,41 @@
+DROP TABLE IF EXISTS tUser;
+CREATE TABLE tUser (
```
dzacharakis, line 2: Why table `tUser` does not have password column but your model has it?

```go
type User struct {
	Id       string `json:"user_id"`  // user id is actually the keycloak user id
	Username string `json:"username"` // this should be the keycloak username
	Password string `json:"password"`
	Email    string `json:"email"`
}
```

server/userHandler.go
vpagal: That is a very good question since I have already implemented the relevant method in userRepository.go, but forgot to push the change removing the stub :(

load_initial_data.sql
vpagal: I did it this way as I had in mind an integration with keycloak. So for user creation I would need the password and via Keycloak REST Api I would create a user there, but for the user data required in our DB password field is not needed.
In our DB we will need a user entry in order to associate this with the Keycloak entity.

At least that was my initial idea about the user management approach for this case, of course there might be better approaches :)

dzacharakis:
Thanks again @vpagal for your time answering my questions and the effort!

PR #9, opened 2023-05-02
GWI go-platform-challenge submission

christosnt:


README.md
```diff
@@ -1,31 +1,184 @@
...
+
+`DELETE /users/{user_id}/delete`	*Delete a user manually (**requires privileged user**)*
+
+  
+#### Cache
+
+This implementation can support any type of cache service by utilizing the created `CacheHandler` interface and implementing its methods.
+The results of a *GET* request to fetch the favourite assets list of a user are cached for faster future retrieval. When this list is updated (either by adding or removing assets), the cache entry is invalidated and the next time the same *GET* request is triggered, the updated resuls will be cached anew. Cache utilizes a map using the client cookies as keys and the *GET* response as the value.
+__At this time a resizing function for these maps is not implemented. To prevent an ever-growing memory allocation (since maps don't shrink) we can either create a fixed capacity map and handle the scaling (up and down when needed) manually, or create a map without specifying capacity and on events (intervals or triggers) scale the map down by recreating from scratch and adding the existing keys to it while setting the previous map to *nil* and allowing the GC to clean it up.__
```
lkslts64, line 94: Good point about the cache size! You mention two solutions for resetting the cache's capacity. Can you provide an example use case where the first one is more suitable and an example where the second one is preferable?

webserver/server.go
```diff
@@ -0,0 +1,63 @@
...
+)
+
+type WebServer struct {
+	Server        mux.Router
+	Configuration *cfg.Configuration
+}
+
+// CreateServer create a new Server instance.
+// Optionally, it accepts options for mocking purposes
+func CreateServer(options ...string) (*WebServer, error) {
```
lkslts64, line 26: How about accepting `*cfg.Configuration` instead of `...string` as the argument of the above function? Would that work?

storage/memoryHandler.go
```diff
@@ -0,0 +1,314 @@
...
+
+	"golang.org/x/crypto/bcrypt"
+)
+
+// memoryHandler is the access handler for the in-memory storage option
+type memoryHandler struct {
+	Block    map[string]map[string]models.Asset // UserID: AssetID: Asset{}
+	Users    map[string]models.User
+	Released bool
+}
```
lkslts64, line 17: Is your storage layer safe for concurrent use? If not, how could you make it thread-safe?

cache/memCache.go
```diff
@@ -0,0 +1,116 @@
...
+		Response:  response,
+		Set:       true,
+		ExpiresAt: time.Now().Add(m.DefaultCacheExpiry),
+	}
+	if len(expiration) > 0 {
+		newCacheEntry.ExpiresAt = expiration[0]
+	}
+	// Start the invalidator counter (if requested)
+	if invalidator {
+		go m.cacheEntryInvalidator(&newCacheEntry)
```
lkslts64, line 60: Are the goroutines spawned by this statement bounded or can they grow uncontrollably? How can you mitigate such an issue?

middleware/prometheusMetrics.go
```diff
@@ -0,0 +1,27 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+var (
+	httpDuration = promauto.NewHistogramVec(prometheus.HistogramOpts{
+		Name: "webserver_http_duration_sec",
+		Help: "Duration of HTTP requests in seconds.",
+	}, []string{"path"})
+)
```
lkslts64, line 16: Nice one!

models/models.go
```diff
@@ -0,0 +1,72 @@
...
+	"github.com/golang-jwt/jwt/v4"
+)
+
+// Asset represents an asset in our system.
+type Asset struct {
+	ID          string `json:"id"`
+	Type        string `json:"type"`
+	Description string `json:"description"`
+	// AssetData contains the asset specific data for Chart/Insight/Audience.
+	AssetData json.RawMessage `json:"assetData"`
```
lkslts64, line 16: Let's say you have a request like `POST /users/{user_id}/favourites` that accepts an asset as part of its payload. If you had to validate that the `AssetData` JSON payload adheres to the schema defined by one of `Chart`, `Insight` or `Audience` types, how would you do that?

lkslts64:
Hey @christosnt thanks for submitting the challenge! Could you please answer my questions in the comments above?

README.md
christosnt: Typically, a fixed capacity map (cache) would be used if we had apriori knowledge of the amount of data that can be stored (in this case number of users) or if we had constraints (memory size in this case). Additionally, the fixed size cache (an array in this case) provides faster operation times (in general). 

A dynamically allocated cache would be useful in a scenario where the resources are elastic (meaning they can be increased/decreased on demand for example for cost saving) as well as when we have no prior knowledge regarding the amount of data that will be cached.

storage/memoryHandler.go
christosnt: In this case it is not. I took a very simple approach regarding user access on the storage probably making assumptions on access patterns that in hindsight might not be safe. It is simple to make it thread-safe by utilizing a Mutex on the Storage. The most safe approach would be for the lock to be requested on all operations, both Write and Read.

webserver/server.go
christosnt: Yes it would work, provided that modification are made to hande the passed Configuration struct ofcouse.

cache/memCache.go
christosnt: In this case they are not bound. They have two termination clauses (memory release, expiration) but manual setting of no expiration or a massive influx of new cache entry requests could impact performance. This can be mitigated by either having a single invalidator doing all the checks (various methods of implementation - brute force, channel messages, etc) which could lead to a delta between expiration time and eviction, or by having a set amount of goroutines performing this operation (tasks on buffered channels, etc).

models/models.go
christosnt: We can either use an external package (validator) to handle the validation or we can do the validation manually by trying to Unmarshall the AssetData to the individual structs (Chart, Insight, Audience).

christosnt:
> Hey @christosnt thanks for submitting the challenge! Could you please answer my questions in the comments above?

I hope i covered your questions. Let me know if you have any more questions.

models/models.go
```diff
@@ -0,0 +1,72 @@
+package models
```
dzacharakis, line 1: I know that was optional, but for a web application it's the data it presents and stores that matters. If you're going to add data persistence to this application what would have used? Why?

models/models.go
```diff
@@ -0,0 +1,72 @@
...
+
```
dzacharakis, line 2: I'm sensing a bottom-up approach in design -as an analyst, you inspected all the interfaces that the system needs to have and worked backwards through the system to determine what and how data should be stored. What do you think, is well suited?

models/models.go
```diff
@@ -0,0 +1,72 @@
...
+import (
+	"encoding/json"
+	"time"
+
+	"github.com/golang-jwt/jwt/v4"
+)
+
```
dzacharakis, line 9: For a database-backed web application, where data is persisted to a relational database, what are the features of a good design? How is it accomplished?

models/models.go
```diff
@@ -0,0 +1,72 @@
...
+type Chart struct {
+	Title      string `json:"title"`
+	XAxisTitle string `json:"xAxisTitle"`
+	YAxisTitle string `json:"yAxisTitle"`
+	// Data type is `any` to handle any format of chart data representation.
+	Data any `json:"data"`
+}
+
+// Insight defines the properties of an insight asset.
+type Insight struct {
```
dzacharakis, line 31: In the design process user requirements are translated into a conceptual schema of a choosen data model, describing the data and their relationships. Do you think `Insight` satisfies the requirements conceptually?

models/models.go
```diff
@@ -0,0 +1,72 @@
...
+	Text string `json:"text"`
+}
+
+// Audience defines the properties of an audience asset.
+type Audience struct {
```
dzacharakis, line 36: Do you think `Audience` satisfies the requirements conceptually?

dzacharakis (commented):
Thank you for your submission @christosnt ! We really appreciate the time you took for this. The lack of data persistence made me want to cover this area, can you please also answer these?

Thanks

models/models.go
christosnt: With the current problem statement i don't see any clear indicators that would promote the use of SQL over NoSQL database or vice versa. With this in mind, i would probably go with a NoSQL database (e.g.MongoDB) to allow for schema flexibility (in regards to assets) and scalability. Optimally, i would use an in-memory db (like memcached or mongodb as in-memory db) and periodically dump it to disk storage. In this way we can have the best performance. It opens up possible issues that can occur because of its volatility but there are ways around them (periodic dumping to disk, enhanced use of cache for changes, etc...).

models/models.go
christosnt: The implementation definitely has the marks of a bottom-up approach. It made sense at the time since this is  the "total" image of the system/platform that was provided in the description. Assumption could be made regarding the design and integration with other possible components but assumptions also have a 50/50 chance to increase the delta with the possible actual request. This design makes sense in the worldview that it is somewhat a standalone (as much as it can be) component, but in a bigger architecture with possibly many components, microservices, etc... total or partial redesign/extention/modification is certainly possible.

models/models.go
christosnt: There are a few features that comprise a good design. Morevoer, they can be made more speficic given the relational nature of the utilized database. In general these features are : scalability, maintainability, reliability, security and ofcourse performance. 

I'll describe what these features encompass in the context of a relational database. 
- Scalability is supported by creating efficient database schemas, using appropriate indices and where/when needed using advanced scaling techniques such as sharding. 
- Maintainability is the ease with which the application can be updated to include new features or if requirements change. This relies heavily on good code organization through the use of clear design patterns.
- Reliability is focused around making sure that the application does not have unexpected behaviour (crash, freeze, unhandled events, etc...) as well as preventing any data loss. This feature is best addressed on both the application and the platform level.
- Security should be one of the main focus points of the application, especially when dealing with user data. Even more so if the data are extremely sensitive (financial, medical, etc...). Nowadays, the bare minimum should be authorization,authentication and encryption mechanisms.
- Performance of the application is one of the most coveted targets and its importance is exacerbated when dealing with massive amount of users. Typical optimization techniques that would lead to low response times in this case are optimized queries to the database and use of caching to reduce database operations.

The typical steps to accomplishing all these features, and in the desired degree, include: 
- a clear definition of the application's requirements
- definition of the database schema (minimizing as much as possible redundancy and utilizing as much as possible relationships) and creation of the database taking into consideration security aspects.
- Develop the application making use of appropriate design patterns.
- Optimize the db queries that the application makes.
- Examine the application's behaviour in order to detect optimization opportunities. This is typically a combination of profiling and monitoring.
- Make sure that security consideration are honored in all cases. A part of the security/reliability considerations should always be a backup/recovery strategy.

models/models.go
christosnt: I think,in this case, it satisfies the requirements regarding the data but not their relationships.

models/models.go
christosnt: The answer is the same above for the Insight. The data requirements are satisfied but not the relationship requirements.

christosnt:
I hope i covered your questions. Let me know if you have any more questions.

models/models.go
dzacharakis: No worries, I have nothing against your decision, but we mainly use databases, so I wanted to hear more about SQL dialects that you might know, how you used to utilize databases (with raw SQL or ORMs), and patterns in order to have a better understanding about your SQL knowledge.

models/models.go
dzacharakis: Nice! And if we reduce the scope of the question to just the problem of designing a schema for a relational database, the goal is to store information without unnecessary redundancy, yet also to retrieve information easily. Any formal approach?

models/models.go
dzacharakis: Yes, this looks like a dangling entity.

models/models.go
christosnt: I see. Regarding SQL dialects, more recently i have used PL/Pg through the use of PostgreSQL. Since most of the applications that utilized this database were applications that collected/analyzed/vectorized metrics from a large amount of devices/sensors/probes/etc so they inlcuded a very large amount of operations we wanted to optimize performance as much as possible wherever possible. This was the main reason for opting to go with raw SQL queries instead of ORMs. Not that i am against them but i prefer the versatility and the optimization opportunities that the raw queries bring to the table. Regarding patterns, nothing too overcomplicated, typical 1:1 and 1:N relationships.

models/models.go
christosnt: I guess that would be normalization. Now, which form to follow though would be dependent on the requirements, while keeping in mind that higher forms lead to higher db complexity thus impacting the maintainability i mentioned earlier (and possibly the performance).

README.md
lkslts64: Makes sense! Thanks @christosnt

storage/memoryHandler.go
lkslts64: Good job! The mutex approach will do the work 🚀

cache/memCache.go
lkslts64: Thanks, mate! That was exactly what I was looking for 🚀

models/models.go
lkslts64: 🚀

lkslts64:
Hey @christosnt, thanks for answering all my questions! 🚀

dzacharakis:
> I hope i covered your questions. Let me know if you have any more questions.

I'm fine, thanks again @christosnt for your time!

main.go
```diff
@@ -0,0 +1,26 @@
+package main
+
+import (
+	"challenge/middleware"
+	"challenge/webserver"
+	"log"
+	"net/http"
+)
+
+func main() {
```
lunemec, line 10: Do you know/have you used tools that help you with structuring CLI apps that help you with ENV parsing, config read/parse and so on?

webserver/handlers.go
```diff
@@ -0,0 +1,501 @@
...
+
+// SigninExistingUser logs in existing user
+func SigninExistingUser(w http.ResponseWriter, r *http.Request) {
+	var existingUser models.User
+	err := json.NewDecoder(r.Body).Decode(&existingUser)
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		fmt.Fprintf(w, "Error decoding request body")
+		return
+	}
```
lunemec, line 29: This part repeats everywhere in all handlers and middleware, how would you approach in improving this? Also this is very prone to `WriteHeader called multiple times` error. Ideas?

webserver/routes.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+func RegisterRoutes(srv *WebServer) { // router *mux.Router
+	if srv.Configuration.Metrics {
+		srv.Server.Path("/metrics").Handler(promhttp.Handler())
+	}
+	// Register the endpoints
+	srv.Server.HandleFunc("/signup", CreateUserHandler).Methods("POST")
```
lunemec, line 15: What frameworks have you used?

main.go
christosnt: I am aware of Cobra for structuring/handling CLI apps. I've mainly used Viper for env/config reading/parsing for multiple components i had to develop.

webserver/handlers.go
christosnt: Of course this code can be replaced by a function that that does the same exact thing. This would not probably address the issue with the error you are referring though. A possible solution would be to handle this functionality, by a middleware that adds it on all requests.

webserver/routes.go
christosnt: I typically use Gin, but decided to try out Gorilla for this assignment given the opportunity.

christosnt:
I hope i covered your questions. Let me know if you have any more questions.

PR #10, opened 2023-05-09
Adds user asset service

loukaspe:


examples/getAssets.http
```diff
@@ -0,0 +1,5 @@
+GET http://localhost:8080/users/1/favourites
+Accept: application/json
+Authorization: Bearer eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhIiwiZXhwIjoxNjgzNDkzMTk3LCJVc2VySW5mbyI6eyJVc2VybmFtZSI6ImEiLCJQYXNzd29yZCI6ImIiLCJGYXZvdXJpdGVBc3NldHMiOm51bGx9fQ.pE5k4PEh7SkTVvPgmcNpvcfuL2NOmilid2gERlatYT8
+
+###
```
dzacharakis, line 5: Nice having the raw HTTP requests. Could you provide an example of usage maybe of a tool (HTTP client) using those files?

README.md
```diff
@@ -1,31 +1,93 @@
...
+## Performance
+
+If Users favourites data were causing performance issues, I would investigate the following solutions:
+
+1. Add a cache to the repository layer, over the DB and keep there the User's favourite Assets
+   in order to have faster access to it
+1. Limit Payload by implementing Pagination
+1. Limit Payload by compressing the data with Accept-Encoding: gzip or compress
+1. Add indexes in DB. For example we could add indexed to the Users-Assets tables e.g "users_charts" on user_id
+   and chart_id
```
lkslts64, line 86: Good job mate! I like the fact that you proactively added those before we even ask you 👍

README.md
```diff
@@ -1,31 +1,93 @@
...
+* All these are executed through docker containers
+* In order to execute makefile commands type **make** plus a command from the table above
+
+  make {command}
+
+---
+
+## Notes
+
+1. .env is pushed to Git only for the Assessment purpose. Config and .env files should never be tracked.
```
lkslts64, line 60: In your view, how config and secrets should be managed? If they shouldn't be tracked by version control, how could someone share them with other collaborators?

internal/repositories/userRepositoryDelete.go
```diff
@@ -0,0 +1,33 @@
...
+	"gorm.io/gorm"
+	"net/http"
+	"strconv"
+)
+
+func (repo *UserRepository) DeleteUser(ctx context.Context, uid uint32) error {
+	// I've chosen to do a soft delete on Users as supported by Gorm. On many platforms
+	// when deleting a user, he is not actually deleted and can restore his account.
+	// This is done by keeping user-asset relationships and delete by updating the deleted_at
+	// field in the DB
```
lkslts64, line 16: Good job! I like this one 💯

internal/handlers/userFavourites/addHandler.go
```diff
@@ -0,0 +1,168 @@
...
+//			$ref: "#/definitions/AddUserFavouriteAssetResponse"
+//	"404":
+//		description: Requested user or asset not found
+//		schema:
+//			$ref: "#/definitions/AddUserFavouriteAssetResponse"
+//	"500":
+//		description: Internal server error - check logs for details
+//		schema:
+//			$ref: "#/definitions/AddUserFavouriteAssetResponse"
+func (handler *AddUserFavouriteHandler) AddUserFavouriteAssetController(w http.ResponseWriter, r *http.Request) {
```
lkslts64, line 66: Why you opted for a pointer receiver in this method?

README.md
```diff
@@ -1,31 +1,93 @@
...
+2. There are three Dockerfile files.
+    1. Dockerfile is the normal, production one
+    2. Dockerfile.dev is for setting up a remote debugger Delve
```
lkslts64, line 63: What are the use cases for a remote debugger?

internal/repositories/userRepositoryGet.go
```diff
@@ -0,0 +1,116 @@
...
+				Title:       chart.Title,
+				XAxisTitle:  chart.XTitle,
+				YAxisTitle:  chart.YTitle,
+				Data:        chart.Data,
+				Description: chart.Description,
+				AssetType:   "chart",
+			})
+			mutex.Unlock()
+		}
+	}(&modelUser.FavouriteCharts, &favouriteAssets, &mutex)
```
lkslts64, line 64: Good job on making this piece of code concurrent 👍  One thing you might have heard is that parallelizing a piece of code doesn't guarantee performance speed up. In fact, the opposite may happen. What are the reasons for that? Is there any chance your "parallel" implementation is slower than the "sequential" equivalent? If yes, how would you benchmark it?

lkslts64 (commented):
Thanks for submitting the challenge @loukaspe! Great work 👍 Could you please answer my questions in the comments?

README.md
loukaspe: Thank you! The assignment was asking specifically to keep an eye on performance but I did not provide any special implementation.The reason is that I believe performance solutions should be given after examining what's causing the delay and only if a performance problem arise. Usually if you try to optimize, without existing need for optimization, the code turns very "smelly".  
After identifying the bottleneck, we could check possible solutions and choose the most appropriate one. The ones I gave are the most common and simple ones that I would start checking.

README.md
loukaspe: Config and .env files usually contain sensitive data that we would not want tracked and seen by everyone in any case. 
Firsty, for our convenience, we could have a dummy file, for example `.env.test`, with dummy values inside, that would be tracked and used for specific reasons (for example integration tests). Or a `.env.local` for the local environment, again with dummy values. 
Anything containing secrets or sensitive data could be in a Password-Secret management tool like Keeper, or Last Pass. Then our collaborators would be given access only to the records that concern them.  
In deployments, the env/config files could be loaded automatically through CI/CD pipelines so that no one has actual access or view to them.

internal/handlers/userFavourites/addHandler.go
loukaspe: This one has been made by accident. There is no specific reason for me doing it. The only thing I know I should be careful about is to have either pointer or value receiver to all the methods and not a mixture of them.

README.md
loukaspe: A use case I have used Delve for is to set up the debugger for a Docker running Go app, either local or in AWS instance for example, and then be able to set breakpoints and debug through my IDE.

internal/repositories/userRepositoryGet.go
loukaspe: Yes, indeed parallelizing could end up in slower speed. First reason could be bad synchonization in the code between go routines that end up in slower performance. But even if the synchronizing is done correctly by the engineer, synchronizing could create an overhead and cause goroutines to wait for each other. Secondly, a not balanced distribution of load could result in bad performance. For example in our problem, to have 100 assets with 98 charts, 1 insight and 1 audience.

I would test it in two ways. Firstly, write benchmarks unit tests for this function, using Go's `func BenchmarkXxx(*testing.B)` tests.
Secondly with actual integration tests and/or UAT to check how it performs. If the performance is acceptable, then we're OK. If not, then I would do the same tests for both solutions and choose the better one.

internal/core/domain/asset.go
```diff
@@ -0,0 +1,14 @@
+package domain
```
dzacharakis, line 1: `platform-go-challenge/internal/core/domain` package is placed in a subdirectory of a directory named internal. Is this just a convention of a standard go project layout like (/build, /cmd, /config, /deployments, /docs, /examples, etc)?

internal/core/domain/user.go
```diff
@@ -0,0 +1,24 @@
+package domain
+
+import "golang.org/x/crypto/bcrypt"
```
dzacharakis, line 3: Nice using `crypto/bcrypt` package for passwords. 👍

internal/repositories/userAssetRepository.go
```diff
@@ -0,0 +1,407 @@
...
+
+	err = repo.db.WithContext(ctx).
+		Preload("FavouriteCharts").
+		Model(User{}).
+		Where("id = ?", userId).
+		Take(&modelUser).Error
+
+	if err == gorm.ErrRecordNotFound {
+		return apierrors.UserNotFoundErrorWrapper{
+			ReturnedStatusCode: http.StatusNotFound,
```
dzacharakis, line 51: In the repository that is the lowest level, you return HTTP response status code that is another level's logic. What do you think about it?

internal/repositories/dao.go
```diff
@@ -0,0 +1,31 @@
+package repositories
+
+import "gorm.io/gorm"
+
+type Chart struct {
```
dzacharakis, line 5: Nice having DAO layer as the abstraction of data persistence 👍

internal/repositories/dao.go
```diff
@@ -0,0 +1,31 @@
...
+	gorm.Model
+	Title       string `gorm:"not null;"`
+	XTitle      string `gorm:"not null;"`
+	YTitle      string `gorm:"not null;"`
+	Data        string `gorm:"not null;"`
+	Description string `gorm:"not null;"`
+	Users       []User `gorm:"many2many:users_charts;"`
```
dzacharakis, line 12: You defined a many-to-many relationship between `users` and `charts`, a user is associated with zero or more charts and a chart, to zero or more users. I see that you decided to advertise both sides of this relationship to the ORM, by creating navigational properties on both `User` and `Chart` DAOs. Why?

internal/repositories/userAssetRepository.go
```diff
@@ -0,0 +1,407 @@
...
+			OriginalError:      errors.New("userId " + strconv.Itoa(int(userId)) + " not found"),
+		}
+	}
+
+	if err != nil {
+		return err
+	}
+
+	insightAssociation := repo.db.WithContext(ctx).
+		Model(&user).Association("FavouriteInsights")
```
dzacharakis, line 253: Why not in one line?
```go
repo.db.WithContext(ctx).Model(&user).Association("FavouriteInsights").Delete(&Insight{Model: gorm.Model{ID: uint(assetId)}})
```

internal/repositories/userDao.go
```diff
@@ -0,0 +1,12 @@
...
+
+import "gorm.io/gorm"
+
+type User struct {
+	gorm.Model
+	Username           string     `gorm:"not null;"`
+	Password           string     `gorm:"not null;"`
+	FavouriteCharts    []Chart    `gorm:"many2many:users_charts;"`
+	FavouriteInsights  []Insight  `gorm:"many2many:users_insights;"`
+	FavouriteAudiences []Audience `gorm:"many2many:users_audiences;"`
```
dzacharakis, line 11: ORM will create three join tables, so your relational schema contains these: `users_charts`, `users_insights` and `users_audiences`. However, semantically you can think of this relationship, favourite, between an asset and a user. In `platform-go-challenge/internal/core/domain` package you defined `Asset` as an interface type, and `Chart`, `Audience`, `Insight` concrete types that satisfy it. This expresses generalization that lacks from the relational schema causing redundancy and cost (three tables, and three joins are needed to get the user's favourites). 

If you had implemented this approach, how would have solved this?

internal/repositories/userRepositoryGet.go
```diff
@@ -0,0 +1,116 @@
...
+) (*domain.User, error) {
+	var err error
+	var modelUser *User
+	var mutex sync.Mutex
+	var wg sync.WaitGroup
+
+	err = repo.db.WithContext(ctx).
+		Preload("FavouriteCharts").
+		Preload("FavouriteAudiences").
+		Preload("FavouriteInsights").
```
dzacharakis, line 26: The cost I was talking about.

internal/handlers/userFavourites/dto.go
```diff
@@ -0,0 +1,79 @@
+package userFavourite
+
+import "github.com/loukaspe/platform-go-challenge/internal/core/domain"
+
+// request for adding asset to user's favourite <chart|audience|insights>
+//
+// swagger:parameters addUserFavouriteAsset
+type AddUserFavouriteAssetRequest struct {
```
dzacharakis, line 8: Nice having DTOs 👍

pkg/server/routes.go
```diff
@@ -0,0 +1,47 @@
...
+	userRepository := repositories.NewUserRepository(s.DB)
+	userFavouriteService := services.NewUserFavouriteService(s.logger, userRepository)
+	assetTypeValidator := helpers.NewAssetTypeValidator()
+
+	createUserFavouriteHandler := userFavourite.NewAddUserFavouriteHandler(userFavouriteService, assetTypeValidator, s.logger)
+	getUserFavouriteHandler := userFavourite.NewGetUserFavouriteHandler(userFavouriteService, s.logger)
+	updateUserFavouriteHandler := userFavourite.NewUpdateUserFavouriteHandler(userFavouriteService, assetTypeValidator, s.logger)
+	deleteUserFavouriteHandler := userFavourite.NewDeleteUserFavouriteHandler(userFavouriteService, assetTypeValidator, s.logger)
+
+	protected.HandleFunc("/users/{user_id:[0-9]+}/favourites", createUserFavouriteHandler.AddUserFavouriteAssetController).Methods("POST")
```
dzacharakis, line 43: Nice using the `gorilla/mux` package's `HandleFunc` with regex pattern 👍

dzacharakis (commented):
Thank you for your submission @loukaspe! We really appreciate the time you took for this. Can you please also answer to me the following questions? 

Thanks

examples/getAssets.http
loukaspe: I'm a fan and i'm using Postman as a tool, it has many functionalities and great UI. I could export my collection there and send it but I didn't want to enforce it to you. That's why I chose simple and raw HTTP requests,  in order to use them in a tool of your preference.
The file as it is used also in JetBrains IDEs, like Goland.

internal/core/domain/asset.go
loukaspe: Lately, my main concern in my APIs is the separation of layers. With that in mind I found the Hexagonal Architecture, which I try get better at, by using it in my last projects. This architecture has the advantage that it enforces you (in a way) to separate the app layers and extract business logic from the technology oriented parts of the app (e.g. PostgreSQL in the repo layer or http in the presentation layer). Here's the article that I found the most comprehensive about the subject:
https://medium.com/@matiasvarela/hexagonal-architecture-in-go-cfd4e436faa3.
That's where the internal dir came from. The rest of the directories do come from the standard go project layout.

internal/repositories/userAssetRepository.go
loukaspe: It's a big mistake that I had not observed. Great review, thank you. The reason I created this kind of errors was to separate the data-info that the client of the API sees (which maybe I want to hide from the client for example) VS the actual data-info that I need in my logs. But I did it in a wrong way of course. I mixed my layers and broke the technology agnostic logic of the Hexagonal Architecture. My bad.

internal/repositories/userAssetRepository.go
loukaspe: I use visual guides in my IDE (80 characters per line) in order to have two open files in one screen without need of horizontal scrolling :P
It's just my way of personal coding.

internal/repositories/userDao.go
loukaspe: At first that was my intention. To generalize both the domain and the repo layer. I had designed everything I wanted on paper and started implementing using gorm. But gorm was actually the one that did not let me use an Asset interface approach there, due to it's design (maybe it was possible, but I could not do it in that time). So due to the deadline, I changed my approach and just made it work the best way possible in the repository layer. And thankfully the domain, which I use in the rest of the app, remained generalized.
My intention was a schema like 

_assets {
  id
  type
  description
}

charts (
  id
  ...
  asset_id
);

insights (
  id
  ...
  asset_id
);

audiences (
  id 
  ...
  asset_id
);

 users {
  id
  username
  password
}  

user_favorites {
  user_id
  asset_id
}_
, but I could not modelize it using gorm. And changing gorm would be too "expensive" and time-consuming for the deadline.

pkg/server/routes.go
loukaspe: Thank you! I must get used to another router package though, as gorilla/mux is in maintenance mode now :(

internal/repositories/dao.go
loukaspe: In my design and my modelization, I wanted User and Assets, to have all the fields-members that the problem required mentally. For example, a User has many charts and a Chart can be a favourite of many Users. I wanted my modelization to show this and to be understandable by reading the code. 
The actual problem solution did not require this though, and that's why in READ operation of an Asset the Users are not preloaded. In order to avoid the extra cost, since I'm using this.
I thought it was a balanced win-win between the readability and the modelization of the problem and the actual performance that would be needed for loading Users from an Asset.

examples/getAssets.http
dzacharakis: Nice, I'm also a big fan of Postman! However, I'm also a big fan of terminal and I thought you used raw data with low level tool like `telnet` or `netcat` to create a TCP connection on the HTTP server.

internal/core/domain/asset.go
dzacharakis: Great about the separation of layers, however, the internal directory is not just a directory. As of Go 1.4, the internal package is a mechanism. Go has only two forms of access: local, visible only within the same package, and global, visible to the world. Sometimes, though, something in the middle would be helpful, that's where the internal package mechanism comes in.

internal/repositories/userDao.go
dzacharakis: > My intention was a schema like

Yeah, to put parent and children in separate tables enforcing parent-level keys. Nice.

> So due to the deadline, I changed my approach

When you are using object-relational mappers you must create formal mapping instructions telling the ORM how to map your entities. I don't know particularly in GORM but on other ORMs like Hibernate, it uses special attributes exclusively for schema generation, so you have no need to specify if schema generation is disabled (the default).

>  but I could not modelize it using gorm. And changing gorm would be too "expensive" and time-consuming for the deadline.

Why didn't you write the DDL SQL first and then map your entities with the ORM? 

Do you consider schema generation as best practice for a production environment?

lkslts64:
Thanks for answering my questions @loukaspe. Great work 👍

internal/repositories/userDao.go
loukaspe: As I said, in the point when I decided to change schema, it was all about the deadline and time management.
The easiest, quickest and simplest choice was to use gorm schema generation and adapt my logic to it. It was a difficult trade-off, and maybe a wrong one. 
I certainly wouldn't use it in a production environment. In an app design for production, the DB schema, as the base of the system, would be a priority and I would want absolute control over the exact structure of the database.

internal/core/domain/asset.go
loukaspe: Did not know that, a very useful and cool feature. Thank you for the  knowledge sharing !

script/Makefile
```diff
@@ -0,0 +1,89 @@
+generate-mock:
```
teliaz, line 1: Kudos mate very clean makefile. Why didn't you place this file in root path?

pkg/server/server.go
```diff
@@ -0,0 +1,65 @@
...
+	go func() {
+		if err := s.httpServer.ListenAndServe(); err != nil &&
+			!errors.Is(err, http.ErrServerClosed) {
+			log.Fatal(err)
+		}
+	}()
+
+	quit := make(chan os.Signal, 1)
+	signal.Notify(quit, os.Interrupt, os.Kill, syscall.SIGTERM)
+	<-quit
```
teliaz, line 51: Can you describe the purpose of that section?

build/Dockerfile.dev
```diff
@@ -0,0 +1,26 @@
+FROM golang:1.19.3-alpine3.16 as builder
+
+WORKDIR /app
+
+COPY ../go.mod go.sum ./
+RUN go mod download
+RUN go install github.com/joho/godotenv/cmd/godotenv@v1.4.0
+RUN go install github.com/go-delve/delve/cmd/dlv@latest
+
+FROM golang:1.19.3-alpine3.16
```
teliaz, line 10: Can you name a few of the benefits of a multistage build?

internal/handlers/userFavourites/dto.go
teliaz: Indeed. Although in a larger project how would you structure those files?

examples/addAsset.http
```diff
@@ -0,0 +1,10 @@
+POST http://localhost:8080/users/1/favourites
```
teliaz, line 1: Thanks for providing those. Are these consumable from Thunder Client?

teliaz (commented):
Thank you for submitting the test mate! Can you answer the following alongside my comments? 

Thanks in advance and again thanks for the effort you put in this assignment!

loukaspe:
First of all, thank you all for your reviews and the opportunity you give me.

@teliaz I will answer here your questions because I can't find a way to reply to your comment.

**Are you following a specific design pattern on you submission?**

No, in this CRUD I did not apply any design patterns. I did not find any good uses for the patterns I'm familiar with.

**How do you value code comments and when do you think that they are necessary?**

I believe sometimes comments are necessary, though must only be used in very specific occasions. I have actually created a guide in my company about Commenting, in which I separate commenting in a greenfield project and in a legacy and in an unmaintanable, poorly-written and designed, error prone system. Here's a quick look : 

----------------------------------

1. Readable, self-explanatory code over commenting. Good comments do not excuse unclear code. Comments should be our last resort.
2. Maintain and update comments
3. Review strictly to ensure that only necessary comments are being accepted​
4. Add comments about TODOs (only when connected with Technical Debt Issues created in a Project Management tool like Jira). Format //TODO: <Issue Code>
5. Add comments that explain the 'Why' of the code. That give context or explain why the code does what it does, that explain decision-making.
6. For very complex code, that is difficult to be self-explanatory, prefer Documentation over Comments

About legacy code we have specifically:
1. Add comments when fixing complex bugs.
2. Add comments about unavoidable complex code.
3. Add comments about hard-coded code (nails) that should explain why they exist

pkg/server/server.go
loukaspe: Sure, this part of code is meant to gracefully shutdown a server after an interruption of OS. It creates and waits on a channel that catches Signals sent by the OS as an interruption (e.g when we press Ctrl + C).
This way, we can then shutdown the server on a way we want (for example server active requests and then close all connections etc)

build/Dockerfile.dev
loukaspe: Yes, of course. Here are a few ones:
1. When deploying a Go app, we only want to run the app. So in that sense, we only need an executable (which is created in build stage) and there is no need to copy any other files.
2. The result of the above is that the image created will be a lot smaller in size than doing the process in only one stage.
3. The image does not contain any other file than the executable, so we encapsulate and hide our app's files, architecture, tools, libraries etc
4.We can separate different uses of the build process by doing it in stages. For example, we can have separate stages for building, testing, linting, deploy etc. This helps us make each of these uses better and maintain each of them in the most appropriate way.

examples/addAsset.http
loukaspe: You're welcome. I haven't used Thunder Client, but as I see it, it can truly consume them. I have provided the simplest form of a request in order to enable you use them with any client you want.

PR #11, opened 2023-05-25
User favourite service

Kercyn:


internal/core/domain/audience.go
```diff
@@ -0,0 +1,47 @@
...
+	return a.Asset.GetType()
+}
+
+func (a Audience) GetDescription() string {
+	return a.Asset.GetDescription()
+}
+
+func (a Audience) GetIsFavourite() bool {
+	return a.Asset.GetIsFavourite()
+}
```
lkslts64, line 47: Are those method definitions necessary given that `Audience` struct has an embedded `Asset` field?

internal/adapter/outbound/mongodb/user_repository.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/Kercyn/crud_template/internal/adapter/outbound/mongodb/dto"
+	"github.com/Kercyn/crud_template/internal/core/domain"
+	port "github.com/Kercyn/crud_template/internal/core/port/outbound"
+	apperror "github.com/Kercyn/crud_template/pkg/error"
+	"go.mongodb.org/mongo-driver/bson"
+	"go.mongodb.org/mongo-driver/mongo"
+)
+
+type MongoUserRepository struct {
+	ctx    context.Context
```
lkslts64, line 17: What is the purpose of `ctx` in this struct?

deployments/development/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN go install github.com/go-delve/delve/cmd/dlv@v1.20.2
+
+COPY . .
+
+RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -gcflags='all=-N -l' -o main ./cmd/http/main.go
+
+EXPOSE 8080
+EXPOSE 40000
+
+CMD ["dlv", "--listen=:40000", "--headless=true", "--accept-multiclient", "--continue=true", "--api-version=2", "exec", "./main"]
```
lkslts64, line 19: Good job with the Dockerfile 👍 

In what cases you may want to run your binary with an attached debugger? Would your Dockerfile be different if it were deployed to a production environment?

lkslts64 (commented):
Thanks for submitting the challenge @Kercyn! Can you please answer the questions below?

What are the reasons for choosing a NoSQL DB over a relational one for this specific project? 

You can find more questions in my review comments. Thanks again!

internal/core/domain/audience.go
Kercyn: I tried to follow an inheritance approach (probably not too successfully), and so made each of the Audience, Insight and BarChart structs be considered "Assets" by implementing the `Asset` interface. Additionally, the `GetIsFavourite` function is called on [internal/service/favourites.go](https://github.com/GlobalWebIndex/platform-go-challenge/pull/11/files#diff-25b4f1f4a85e1682d10e81a582d9c573a66432d32641a61fca98b4b3a5c166e7), line 39.

deployments/development/Dockerfile
Kercyn: This Dockerfile is used for the local development environment, possibly could be used for a staging environment as well. For the production environment, you would want a multi-stage build, in which you would build the executable (without debug info) in an ephemeral "builder" image, then you would copy the executable onto a second image and run it from there. This way, you would end up with a lightweight image containing the executable only, without the source code or other unnecessary files.

internal/adapter/outbound/mongodb/user_repository.go
Kercyn: It allows us to inject information such as timeouts for our database calls, as seen on [internal/app/root.go](https://github.com/GlobalWebIndex/platform-go-challenge/pull/11/files#diff-e13d4e4bf77c274abdef9106db2aa3792527330a3d3a3ddb8f35a84c139a1822), line 17.

Kercyn:
> Thanks for submitting the challenge @Kercyn! Can you please answer the questions below?
> 
> What are the reasons for choosing a NoSQL DB over a relational one for this specific project?
> 
> You can find more questions in my review comments. Thanks again!

I went with a NoSQL DB because we have several objects that are structurally different but conceptually similar. By going with an SQL option, I would have to have the following tables:

- a table for _each_ asset type that would store the type's unique attributes
- a table that would hold the shared attributes
- a table for _each_ asset type that would associate the asset IDs of shared and unique attributes of each asset with a user

Adding a new asset type would also require creating new tables on the database, as well as creating the necessary relationships between the aforementioned tables. The queries required would also be more complex, as there would be a need for multiple joins between tables to get a single user with all their assets.

All the above would add complexity to both the database schema and to the code that would handle the I/O with the database. I just couldn't see any benefit that would justify the added complexity over a NoSQL solution, in which adding a new asset type and handling the I/O is more streamlined. In a NoSQL database, you need only a single adapter struct for the new type, instead of the 3 that would be required for the SQL database, not to mention the added functions needed for CRUD operations.

Thank you for taking the time to review my submission!

lkslts64:
Nice one @Kercyn! Thanks for putting in all this effort!

PR #12, opened 2023-06-13
wip: initial proto, gen openapi, app, service apis

ac-i:
hello 
it is work in progress

there is initial protos, generated openapi, app , service apis (ongoing implementations)
in next days I will finish implementation of services
there are visible coexisting alternative approaches
I hope it will be good for discussion and some benchmarks

you can run it simply from root folder as : go run .
then gRPC (env "SERVER_GRPC_ADDRESS", ":9090") and REST GW ("SERVER_RESTGW_ADDRESS", ":9080") endpoints are active
there is registered grpc.reflection service so you can use gRPC client/gui like Dancheg97.grpc-clicker to check endpoints
or you can use openapi gui to visualise proto/apidocs.swagger.json REST GW endpoints

have fun :)


app/app.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func runApp(ctx context.Context, config *Config) error {
+	app, ctx, err := initApp(ctx, config)
+	if err != nil {
+		return err
+	}
+
+	app.server.Serve(ctx)
+	runtime.Gosched()
```
lunemec: Why? you are already calling gosched inside server.Serve()
Also, why do you call gosched after starting new goroutines?

app/app.go
```diff
@@ -0,0 +1,169 @@
...
+func (app *App) Valid() bool {
+	return app.config.Valid() &&
+		app.auth.Valid() &&
+		app.storage.Valid() &&
+		app.server.Valid()
+}
+
+func (app *App) interrupt(ctx context.Context) {
+	app.chErr = make(chan error, 2)  //nolint:gomnd
+	app.chMsg = make(chan string, 8) //nolint:gomnd
```
lunemec: Whats with these channels? Why? They are not used.
Why are these channels buffer sizes 2 and 8?

app/auth/config.go
```diff
@@ -0,0 +1,29 @@
...
+	PassWord string
+	PassHash string
+}
+
+func NewConfigAuth() *ConfigAuth {
+	return &ConfigAuth{
+		AppRoot: &BasicAuth{
+			UserName: env.Env("AUTH_APPROOT_USERNAME", "app-root"),
+			PassWord: env.Env("AUTH_APPROOT_PASSWORD", ""),
+			PassHash: env.Env("AUTH_APPROOT_PASSHASH", "$2a$08$rvzL6XK2K/UPnK1fkHS3TeDWyc9nbYMAnZBSkiiWCamNilcYxA4jK"),
```
lunemec, line 22: What is this for?

app/client/grpc.go
```diff
@@ -0,0 +1,149 @@
...
+	timeoutDial = 10 * time.Second
+)
+
+type GRPC struct {
+	config         *ConfigClient
+	grpcClientConn *grpc.ClientConn
+	cancel         context.CancelFunc
+	grpcDialOpts   []grpc.DialOption
+	mu             sync.RWMutex
+}
```
lunemec: There is a downside with using this approach - by using shared state struct
where you initialize it with New(), and then before using this, you always have to
call ValidConnection() (or DialContext()). There are however ways to do this differently.
Can you describe a different way to implement this? Can you think of a type-safe way to do it?

app/logs/zerolog-logger.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/rs/zerolog"
+)
+
+// Output duplicates the global logger and sets w as its output.
+func Output(w io.Writer) zerolog.Logger {
```
lunemec, line 14: What does wrapping the base zerolog functions achieve?

app/pki/pki.go
```diff
@@ -0,0 +1,378 @@
...
+func (p *PKI) saveKeyPairCert(casePKI casePKI, certPEMBlock []byte, keyPEMBlock []byte) error {
+	const (
+		permDir fs.FileMode = 0740 // 740 rwxr-----
+		permCer fs.FileMode = 0640 // 640 rw-r-----
+		permKey fs.FileMode = 0600 // 600 rw-------
+	)
+
+	saveDir := filepath.Join("test", "certs", "auto-self")
+	saveCer := filepath.Join(saveDir, fmt.Sprintf("%s-cert.pem", casePKI))
+	saveKey := filepath.Join(saveDir, fmt.Sprintf("%s-key.pem", casePKI))
```
lunemec, line 311: Do you think it is a good idea to use files with microservices? How to do this correctly?

lunemec (commented):
Hello,

we looked at this challenge and have some questions.

Feel free to elaborate further on any of them.

Thank you!

app/app.go
ac-i: It is to control order of execution.  Here it is double check any go func in  inside server.Serve() will be scheduled before next line in app (48+). It can happen that  modifying quickly deep code I switch on/off go func so leaving runtime.Gosched() in main routine can help in these cases. It doesn't block if there is nothing to do and it helps to be more ordered for go func in any

runtime.Gosched() allows to process a go subroutine body before next line in 'parent' routine;
go scheduler behaves in way that it first process code in main routine and any go func is postponed until main riches end or sleep or waiting for chan input. In past I used to think go func behaves unordered but this finding shows they are processed in reverse order like defer

It is very useful for initialising waiting group where wg.Add(1) can be added inside go funct /not before (for example flexible number of started routines). Then adding line runtime.Gosched() before wg.Wait() ensures wg.Add(1) in subroutine is processed before wg.Wait(), (otherwise wg.Wait() passes because scheduler didn't process  wg.Add(1) in subroutine)

```sh
for i := 0; i < N; i++ {
go func() {
	wg.Add(1)
	wg.Done()
}()
}
runtime.Gosched()
wg.Wait()

```

Doe it answer your question?

app/app.go
ac-i: it is only a placeholder in current case but it can be turned to use as alternative communication/control in the code
app scope corresponds to runtime and some functionalities can be attached here for more central or shared usage

app/auth/config.go
ac-i: app/auth - is a starting point for authentication and authorization functionality; checking is not implemented but if required it can be turned on on app/env level;

app/client/grpc.go
ac-i: Do you think about sync.Once?

It could be used in similar functionality.
Here the code is not intuitive enough.  Sequence of staring the server can be improved more. 
DialContext is used once, in fact, by private server client to wait for connection to grpc before starting the rest gw then this private client is not used by other code.
Common usage of grpc in app is based, so far, on  client.InsecureClientConnGRPC() 
which has embedded standard grpc.DialContext (insecure is because pki doesn't use cert in file for tls) 
and it doesn't require sync.Once. 
It can be checked in fake_test.Example_gRPC_Client_loading_fake_data() / grpcClientConn, cancelDial, err := client.InsecureClientConnGRPC(context.Background(), client.NewConfigClient()) / cliU2 := userAPIv2.NewUserServiceClient(grpcClientConn)
repeated rpc calls use one shared grpcClientConn established at once beginning of the example_func as internal var

btw I avoid usage of global variants

Does it answer your question?

app/logs/zerolog-logger.go
ac-i: app/logs is centrally configured structure multi-logger
app/logs/zerolog-logger.go contains helper functions for github.com/rs/zerolog; and they already use initialised configured logger with caller/stack info logging

The func Output(w io.Writer) zerolog.Logger allows to replace output for example to a file. It will work in place of usage only. While app/logs can be configured in one place for all usages
This helper func Output is not used in code but it is here to keep 'standard' of zerolog

app/pki/pki.go
ac-i: It is not the best idea and app/pki doesn't use these files as source but each time overwrite app/pki/test/certs/auto-self. They are to provide content for potential tests of particular case/approach.

How to do this correctly? - It is configured already to use for test purposes on unknown environment with tls switched on with highest crypto standard as possible. Other particular usage depends on the environment and on what we want to achieve. There is no one answer good for all cases.

app/pki simply generates 3 types of key and self sign them.
rest gw and internal cli.grpc use pki.TLSConfigDial() with self-signed cert added to the CertPool() and InsecureSkipVerify is default false so transport is encrypted and server is verified; 
So far they don't use own client's key and server is configured as tls.VerifyClientCertIfGiven - it means server don't verify client;
External grpc cli (like one in fake_test) run on separate os process/instance so it doesn't have the same self-signed certPool and tls config is set to not verify server cert client.InsecureClientConnGRPC() with InsecureSkipVerify: true

It can be enhanced to certify and verify rest gw and client too; 
It can be further modified to certify with external provider (let's encrypt would require external ip) so grpc client in fake could be switched to InsecureSkipVerify: false but it would also require a mechanism of sharing cert or mechanism of certification using cert key corresponding to externally certified pub (last one is recommended)

Entire app is more a study and entry point for discussion or/and tests based on working tool then selected case and path. Also TLS changes dynamically but not all tools uses latest achievements so pki can be used to test periodically where which standard can be switched on.

app/app.go
ac-i: Sizes corresponds to potential purpose/behaviour
They are both buffered to not block place of sending in concurrent usage
chErr - is to decide if to close entire app, it is similar to fatal but logic can be modified in interrupt()
chMsg - is more like conditional logging or other central processing of received info; it's usage can be much higher so size  should help with potential peaks

app/logs/zerolog-logger.go
ac-i: in fact Output can use any io.Writer
In past I had set logging directly to Splunk HTTP Event Collector which used to extract structured logged fields much better then Splunk parsing syslog for the same structures
It can use io.Writer sending log messages to db too ...

app/client/grpc.go
ac-i: btw there is experimental func (cc *ClientConn) GetState() connectivity.State
but it is stil experimental, also using mutex and is  more complex for the purpose

app/app.go
lunemec: OK, thanks. Where would you send data into these channels? Can you provide us with minimal example (just filename/func is enough)?

app/auth/config.go
lunemec: Yes, but I meant the pass hash specifically, why is it in config? What is it used for?

app/app.go
lunemec: Yes, it does. However this is wrong.

There are 2 issues here.

1) you *must not ever* call `wg.Add` from inside the goroutine you are spawning. Because goroutine scheduling is not deterministic and you are NEVER guaranteed that the goroutine you spawned (even when you call gosched) gets executed before you call `wg.Wait`.

Firstly it specifically says so in the docs: https://pkg.go.dev/sync#WaitGroup.Add : ` Add should execute before the statement creating the goroutine or other event to be waited for`

And secondly, this creates a bug that is very very difficult to even find, because in some cases one of your goroutines never gets spawned, but the code looks ok.

2) the runtime.Gosched() should not be used in this type of code, it is used here as a workaround to wrong usage of wg.Add. The main usecases for gosched is for I/O where you can yield your code to allow different goroutines to run. In standard library, the usecases that can be found are: right after calling syscall, signal handling, mutex operations, GC triggers (there are more, feel free to check for yourself: https://cs.opensource.google/search?q=gosched&sq=&ss=go%2Fgo)

app/logs/zerolog-logger.go
lunemec: My question is, why not directly call `github.com/rs/zerolog` `log2C.Output` ? It has the exact same type signature as your `Output()` function.

And on top of that, by creating `LogC2` initialized logger, you can never write tests for logs - you can never swap what you initialized as logger with different logger because it is global to this package.

What I mean by testing logs - for example, we have audit logging in some of our code, where the business requirement is specifically to log certain data after an event. To test your business logic, you have to call the event handling function with a log observer, where you can test for it containing certain data.

app/app.go
ac-i: I agree it is potentially not clear and we should be clear not clever with GO :)

From other hand it doesn't break the docs because it says about execution of Add before the statement (not about placement of statement Add) and we can achieve it in more ways: first clear is to put statement Add before statement go doing add; second is to strictly ensure execution of Add before statement waiting for go statement(s) doing add for example by gosched. And once we have more than two ways we take control on execution of our code. Plus we can suggest an update of documentation ;)

As I had checked the order of execution of go subroutine(s) it is the same as order of execution of defer(s). In first iteration of scheduler they are only initialised with potential arguments. The difference is in triggers where defer is triggered by end of function body or by return while go subroutines are triggered by the scheduler sequentially assigning execution time for main routine and separate subroutines (if main routine not end before).

Example of Add is simple, easy to replace and test each approach, so nice to discuss and learn

Other place where it is used is starting the server where sequence os steps using subroutines is required. It is easy with gosched. Other ways are unequivocal to answer if grpc is listening on net port. Checking too quickly by client can return err. Adding sleep time has more disadvantages then gosched. So with my current knowledge I keep gosched as recommended there for simplicity, control and performance.

app/app.go
ac-i: A trivial example but not recommended can be close in app/app.go in func initApp(
by replacement of some immediate return on err to sending this err to app.chErr and continuing
```go
	// 9
	if err = app.registerServices(app.server.ServiceRegistrar()); err != nil {
		//nolint:dupword
		// return nil, nil, fmt.Errorf("app.registerServices: %w", err)
		app.chErr <- err
	}

then
			case err := <-app.chErr:
				// logs.Error().Err(err).Msg("app.chErr")
				// app.closeApp(ctx)
				app.errResolver(ctx, err)
and

func (app *App) errResolver(ctx context.Context, err error) {
	logs.Error().Err(err).Msg("app.chErr")
	app.closeApp(ctx)
}
``` 

For now both channels are consumed by logs in func (app *App) interrupt(
We can enhance by dedicated functions doing some logic with channel taken as an argument

Intended usage is in subroutines like server - grpc but first these channels must be enabled in these scopes in runtime by app.inst which is populated during initApp. And here we have a hierarchy of components so it will be better to provide only direct parent's channel to child(s) to escalate step by step (not directly to highest level)

I ques it is not the case you are asking for, right?
func (app *App) errResolver( doesn't use chan in arguments nor in return values but it is intentional here.
In case there is required chan in arguments or chan in return I know what it is about and how to work with such cases.

btw 
You probably saw the code is organised in kind of hierarchy and avoids global variables and global exported functions. It is to keep consistency of behaviour with flexibility to change the structure from singular to multiple
for example 
current
type App struct {
	server  *server.Server
can become
type App struct {
	server2  []*server.Server
or
type App struct {
	server3  map[instance.InstMode]*server.Server
where server for now = net listener = port x2 plus registered grpc-service-api & gw-mux

In such case a global or wide scope func working with channels can potentially limit flexibility of change 

Let me know, please, if you wish me to enhance the code to enable channels on service level?

app/auth/config.go
ac-i: _PASSHASH in config is to simplify potential emulation of authenticator to be able to start without external storage.
It is not recommended to save passwords in config so to be able to check if password is env _PASSWORD is the same as hardcoded one there is common methodology to store not password value but secure king of hash of the value like bcrypt.
_PASSHASH is here to distinguish in potential authenticator provided _PASSWORD vs _PASSHASH to be checked against to
If part of functionality depends on authenticated admin~app-root we don't have to relay on external storage but we can validate env _PASSWORD against env _PASSHASH with default value set in code (or set in code as const plus one from env).

Does it answer your question?

lunemec: It does, thank you.

app/app.go
lunemec: Thanks, this is fine. The intent is only to pass errors from goroutines created during initialization. Not in the app logic, that was the goal of the questions. Thank you.

app/app.go
lunemec: > As I had checked the order of execution of go subroutine(s) it is the same as order of execution of defer(s).

Can you please provide your source for this? 

Please note this is irrelevant to the application, I'm just curious, because I don't think this is correct. I looked at this: https://go.dev/src/runtime/proc.go?s=8144:8158 and https://docs.google.com/document/d/1TTj4T2JO42uD5ID9e89oa0sLKhJYD0Y_kqxDv3I3XMw/edit#heading=h.1amka2n5suqo where this one part about scheduling would lead to non-deterministic behaviour depending on each processors list of runnable goroutines:
```
When a new G is created or an existing G becomes runnable, it is pushed onto a list of runnable goroutines of current P. When P finishes executing G, it first tries to pop a G from own list of runnable goroutines; if the list is empty, P chooses a random victim (another P) and tries to steal a half of runnable goroutines from it.
```

Specifically this part:
```
if the list is empty, P chooses a random victim (another P) and tries to steal a half of runnable goroutines from it.
```

So depending on the total number of goroutines in each of processors list of goroutines, it can lead to condition where they would be running in very different order from the code, even after running gosched.

Defers work differently from this, and are always deterministic.

app/logs/zerolog-logger.go
ac-i: app/logs allows central, flexible configuration for all logs
The app/logs is not yet fully integrated because it uses config out of app config and it can be the main source of confusion for now
The intention is to have configurable and consistent behaviour of logs in instance of app. No matter if someone  try to use standard log or app.logs both will go to the same outputs and will have predefined fields entered

To solve your audit logging by app.logs there is required to define one more output  (similar as it was done for Splunk HEC). Then depends if all events should go to this output=observer then it should be added to Logger or if only particular events then we need additional LogX with added output=observer and to use logs.LogX in these particular places (if LogX will be based on Logger then other outputs also receive it's events if not than sending to other will require two loggers in particular places)

app/logs was initially created to solve issue with three separate in-house loggers in one codebase to format and direct all of them to configured outputs (depends on env) and with requirement to not change code in place of usage.  It was really tricky. In comparison your audit logging seams to be very logic and straight to implement.

In app/logs there is preconfigured github.com/rs/zerolog so global defaults of rs/zerolog are changed 
```go
	zerolog.TimeFieldFormat = zerolog.TimeFormatUnixMicro
	zerolog.TimestampFieldName = "µs"
	zerolog.ErrorStackMarshaler = pkgerrors.MarshalStack //nolint:reassign
	zerolog.CallerMarshalFunc = formatCaller
	// zerolog.SetGlobalLevel(zerolog.Disabled)
	// zerolog.SetGlobalLevel(zerolog.DebugLevel)
```
and there are added output(s) and some fields to default Logger
```go
	Logger = zerolog.New(zerolog.MultiLevelWriter(
		c.setSyslog(),
		c.setStdErr(),
		c.setStdOut(),
		c.setConsole(),
	)).
		With().
		Timestamp().
		Str("app", c.App).
		Int("pid", os.Getpid()).
		Logger()
```
and there are separate loggers with different caller
```go
	LogC2 = Logger.With().
		CallerWithSkipFrameCount(2).
```
```go
	LogC3 = Logger.With().
		CallerWithSkipFrameCount(3).
```
and standard log is configured to use Logger output(s) and correct stack caller
```go
	log.SetFlags(0)
	log.SetOutput(Logger.With().
		CallerWithSkipFrameCount(5).
```

Then logs.Output() already uses preconfigured outputs and fields as LogC2
It allows to replace output (if required, but not used in current app)
```go
func Output(w io.Writer) zerolog.Logger {
	return LogC2.Output(w)
```

While zerolog.Output() will add output to global defaults of rs/zerolog
```go
var Logger = zerolog.New(os.Stderr).With().Timestamp().Logger()

func Output(w io.Writer) zerolog.Logger {
	return Logger.Output(w)
```

So, in short, if both are using the same output, there will be difference in format (zerolog globals) and included fields (preconfigured With )

app/app.go
ac-i: I have investigated behaviour playing with code like one below:
add func, trigger, remove=comment, change order ... (below is some state after adding and checking some stats)

The ordered behaviour appears after adding runtime.Gosched() which reverse orders closest go func (like defer)
If runtime.Gosched() is not used go func runs in unpredictable order (not like defer) = like you described above and like I used to think

imo we can switch on order, for at least closest subroutines, by placing runtime.Gosched() otherwise subroutines runs in kind of random

```go
package main

import (
	"fmt"
	"runtime"
	"time"
)

var t time.Time

func say5(s string) {
	// go func() {
	// 	wg.Add(1)
	// 	wg.Done()
	// }()
	// runtime.Gosched()
	// wg.Wait()
	for i := 0; i < 5; i++ {
		t1 := time.Since(t)
		runtime.Gosched()
		t2 := time.Since(t)
		fmt.Println(i+1, s, runtime.NumCPU(), runtime.NumGoroutine(), t1, t2)
	}
}

func main() {
	fmt.Println(runtime.Version())
	t = time.Now()
	fmt.Println("start", runtime.NumCPU(), runtime.NumGoroutine(), time.Since(t))
	go say5("world-sub-1") // create a new goroutine
	go say5("world-sub-2") // create a new goroutine
	fmt.Println("start", runtime.NumCPU(), runtime.NumGoroutine(), time.Since(t))
	// time.Sleep(time.Millisecond)
	say5("hello-main-") // current goroutine
	for i := 0; i < 6; i++ {
		t1 := time.Since(t)
		runtime.Gosched()
		t2 := time.Since(t)
		fmt.Println(i+1, "main-sched-", runtime.NumCPU(), runtime.NumGoroutine(), t1, t2)
	}
	fmt.Println("end", runtime.NumCPU(), runtime.NumGoroutine(), time.Since(t))
	// time.Sleep(time.Millisecond)
	fmt.Println("end", runtime.NumCPU(), runtime.NumGoroutine(), time.Since(t))
}
```

lunemec: Thanks for the example. I still don't think that it will be deterministic 😄 but I don't have any example to prove it. I'll maybe dig into it sometime, but lets leave that for some other time. Thanks.

app/client/grpc.go
ac-i: grpc client and unused code is fixed now and this structure is not in use

If we talk about shared state structures vs mutable data structures vs immutable data structures it all depends on intention
I don't see downsides but rather features, consequences and requirements related to each approach
In the app I focus on zero-copy approach which provides great benefits in performance and of course require careful construction of the code with double-checking if there is no place changing shared value in unintentional way

If there is required immutable approach then we can construct without pointer as
type GRPC struct {
	config         ConfigClient

then use known patterns using getters, setters and withers like:
func (g GRPC) WithConfig(config   ConfigClient) GRPC {
    g.config = config
    return g
func (g GRPC) Config() ConfigClient {
    return g.config

app/app.go
ac-i: I have fixed wg.Add to be more clear and GO way, although discussion was very interesting :)

Also runtime.Gosched() is removed.
In part starting server there is grpc client trying connection before rest gw. It is set to use option grpc.WithBlock(). Usually this option is not required but here is added by intention with purpose to ensure working connection before starting rest gw

app/app.go
ac-i: btw unussed code is now removed or deactivated/commented

lunemec:
@ac-i you can stop adding more stuff now 🙏 😄 I thought someone had contacted you already.

PR #13, opened 2023-07-23
Dev

soring323:
# GWI Testing API

This API was designed for GWI project(testing)

## Running the API

To get the API up and running, follow these steps:

1. **Generate protobuf files**
   Start by generating the protobuf files. Use the following command:

   ```shell
   make generate
   ```

   This will generate the protobuf files and place them in the `./pkg` directory.

2. **Configure JWT**
   Open the `config.env` file located in the `./config/config.env` directory. You will need to add a key-value pair for `jwt.signedKey`. This key can be obtained from OAuth or Firebase, or generated directly using OpenSSL with the following command:

   ```shell
   openssl rand -base64 64
   ```

3. **API Documentation**
   You can explore all available API endpoints using the Swagger UI available at:
   ```
   http://localhost:8901/api/swagger-ui/
   ```
   Please note that the Swagger file is currently not complete, and therefore you should only use it to understand the API request style. After authorization, remember to add the `Authorization` header field with the `Bearer {your token}` to your requests. The token can be obtained during the sign-in or sign-up process.

## Disclaimer

Please note that this API was initially designed for testing purposes and its structure might undergo changes based on more detailed requirements.

For testing purposes, the current version of the API simulates a database in memory instead of using a real database setup. Depending on the finalized requirements, a decision can be made between using a relational or non-relational database in the future.


Makefile
```diff
@@ -0,0 +1,14 @@
+generate:
+	cd api; buf generate; mv ../pkg/microservice.swagger.json ../static/swagger-ui/swagger.yaml
```
jhatzics, line 2: It seems that if the host machine is missing some prerequisite tools this step fails. What would you do to improve the situation here?

docker-compose.yml
```diff
@@ -0,0 +1,21 @@
...
+      - 3306:3306
+      - 8080:8080
+    volumes:
+      # persist data between runs:
+      #- db-data:/var/lib/memsql
+      #- /home/user/init.sql:/init.sql
+      - ./init.sql:/init.sql
+    environment:
+      # use the LICENSE_KEY environment variable set in the terminal:
+      LICENSE_KEY: BDc0N2Y2YTQzZjM2MTQwYTM4MmFlNTJhOGI4Y2I2MGI1AAAAAAAAAAAEAAAAAAAAACgwNAIYMWL4bAtDSFFwpdt5xsVJwuo3/NM6n0ilAhgVC+he8d+tSaNqHwRtQZzvMyDlKwLn6ZYAAA==
```
jhatzics, line 15: How could you avoid submitting those values on git repo?

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM golang:latest
+WORKDIR /app
+ADD . .
+RUN go mod download
+CMD cd cmd && go run main.go
```
jhatzics, line 5: Why are not compiling the app before running it, here? What are the disadvantages of `go run`

cmd/main.go
```diff
@@ -0,0 +1,155 @@
...
+	// JWT
+	signedKeyJWT := viper.Get("jwt.signedKey").(string)
+	tokenManager := service.NewTokenManager(signedKeyJWT)
+
+	// Register all services
+	//dbHandler := repository.NewDBHandler(db)
+	gwiDB := repository.NewDBHandler(gwidb)
+	//logDBHandler := repository.NewDBHandler(logdb)
+
+	userService := service.NewUserService(gwiDB)
```
jhatzics, line 56: Why do you pass all the available handlers (e.g. favourites, asserts) in the user's service? Could you do something better here?

cmd/main.go
jhatzics, line 1: Do you think this file has taken lot of responsibilities? Any suggestions for refactoring?

internal/constants/error.go
```diff
@@ -0,0 +1,12 @@
+package constants
+
+var (
```
jhatzics, line 3: I see those variables not used somewhere in your code? What is their purpose?

internal/dto/user.go
```diff
@@ -0,0 +1,70 @@
...
+		ID:       id,
+		Gender:   gender,
+		Email:    email,
+		Password: password,
+		Country:  country,
+		Age:      age,
+	}, nil
+}
+
+func isValidCountry(country string) bool {
```
jhatzics, line 62: Why this function is not part of a dto and not the user's domain?

internal/repository/db.go
jhatzics, line 1: Do you think this files gathers more than one responsibilities? Any suggestions for refactoring?

internal/repository/favorites.go
```diff
@@ -0,0 +1,23 @@
...
+
+type FavoritesQuery interface {
+	AddFavorites(userId uint64, assetId uint64) error
+	GetFavorites(userId uint64, params domain.PaginationParams) ([]domain.Asset, error)
+}
+
+type favoritesQuery struct{}
+
+func (f *favoritesQuery) AddFavorites(userId uint64, assetId uint64) error {
+	return DB.AddFavorites(userId, assetId)
```
jhatzics, line 17: How could you mock the DB implementation in order, for example, write some unit tests for the repository layer?

pkg/google/api/annotations.swagger.json
jhatzics, line 1: Is it necessary for these generated files (under `/pkg/validate` and `/pkg/google`) to be versioned in git?

jhatzics:
Hello,

First of all, thank you for the effort you put in this assignment. We really appreciate it. Apart from my comments below I would like you to answer the following questions.

### Questions

1. In terms of performance, what would you change in your app before a production deployment? Assume that hundreds of users will use it. Consider that your are using a regular relational DB
2. How would you approach unit testing in your app and why that way? Please be as detailed as possible.
3. Why did you choose to have both gRPC and HTTP interfaces? Do you see any drawbacks executing them from the same golang app?

docker-compose.yml
soring323: that's simple you can use git secret/variable in script. 
very simple things

soring323:
1. for product, it require many changes.
 first of all, we will use cloud database as a single datastore with this model. 
 we will add load balancer to distribute traffic of service. In this case, we can resolve two problem.  In database, with cloud database (AWS Azure, GCP, SingleStore...), we can implement scalability easily 
Second, with load balancer, we can decrease traffic to avoid overwhelming of service. 
2. first of all, we can add unit testing according to the layer. this structure is just clean architecture so can complete according to layer that. Final e2e testing, can be built with GitHub action. 
3. we have 2 options.
Http/GRPC call all is available. 
if we have mobile client or other client, we can use grc call to speed up, and with protobuf, it can be changed it's structure flexibly 
Also add http endpoints for web browse networking.

PR #15, opened 2024-06-05
Implement JWT verification, rate limiting, and handlers with Docker support

meletios:


Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+WORKDIR /app
+
+COPY go.mod go.sum ./
+
+RUN go mod download
+
+COPY . .
+
+ENV JWT_SECRET=your_jwt_secret
```
arsotirchellis, line 11: You are going to pass a secret as an environment variable. How would you approach it in a production system?

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+RUN go test ./...
```
arsotirchellis, line 13: How would you incorporate your tests or your testing suite in general into your daily routine?

README.md
```diff
@@ -1,31 +1,81 @@
...
+
+    ```bash
+    go run main.go
+    ```
+
+4. The application should now be running on `http://localhost:8080`.
+
+## Endpoints
+
+- `POST /login`: Authenticates a user and returns a JWT token.
```
arsotirchellis, line 65: How would you communicate a change in one of these endpoints to the team that consumes your API?

go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/meletios/gwi-engineering-challenge
+
+go 1.22.3
+
+require (
+	github.com/dgrijalva/jwt-go v3.2.0+incompatible
```
arsotirchellis, line 6: Is this library considered safe?

handlers/asset_handler.go
```diff
@@ -0,0 +1,95 @@
...
+	userID := vars["userID"]
+
+	var asset models.Asset
+	err := json.NewDecoder(r.Body).Decode(&asset)
+	if err != nil {
+		http.Error(w, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	UsersFavorites[userID] = append(UsersFavorites[userID], asset)
```
arsotirchellis, line 38: Is this safe?

README.md
```diff
@@ -1,31 +1,81 @@
...
+- `GET /favorites/{userID}`: Retrieves a list of favorite assets for a user.
+- `POST /favorites/{userID}`: Adds a new favorite asset for a user.
+- `DELETE /favorites/{userID}/{assetID}`: Removes a favorite asset for a user.
+- `PUT /favorites/{userID}/{assetID}`: Edits a favorite asset for a user.
```
arsotirchellis, line 69: Is userId safe to be url param ?

handlers/asset_handler.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+	}
+
+	http.Error(w, "Asset not found", http.StatusNotFound)
+}
+
+func EditFavorite(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	userID := vars["userID"]
+	assetID := vars["assetID"]
```
arsotirchellis, line 69: How could a validator be useful here? What types of validations could we do?

handlers/auth_handler.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	tokenString, err := token.SignedString(mySigningKey)
+	if err != nil {
+		http.Error(w, "Could not generate token", http.StatusInternalServerError)
+		return
+	}
+
+	http.SetCookie(w, &http.Cookie{
```
arsotirchellis, line 55: Why did you choose `http.SetCookie` instead of a typical HTTP response?

handlers/asset_handler.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/gorilla/mux"
+	"github.com/meletios/gwi-engineering-challenge/models"
+)
+
+var UsersFavorites = make(map[string][]models.Asset)
```
arsotirchellis, line 11: I believe this database is coupled with this layer. Do you agree with that statement?

arsotirchellis (commented):
Hello!! 
We really appreciated the effort you put into this assignment. 😃
We loved the idea that you added detailed tests and a rate limiter. 
We made some comments to spark some discussions around your implementation. Feel free to respond to some or all of them.

Dockerfile
meletios: Good catch, on way its to use a secret management service such as AWS Secrets Manager or to inject environment variables at runtime using orchestration tools like Kubernetes

Dockerfile
meletios: Using CI tools like Github actions

README.md
meletios: I would update the API documentation, notify the external team, and possibly ask if they would like to setup a  meeting to ensure everyone understands the update

go.mod
meletios: No, I prioritised minimising development time for this assignment over implementing robust security measures, knowing that the code would not be deployed in a production environment.

handlers/asset_handler.go
meletios: No, this is not safe as it lacks concurrency control, which can lead to race conditions in a multi-threaded environment.

README.md
meletios: No, passing userID as a URL parameter can expose it to unauthorised access and potential security risks, its safer to extract it from a secure token

handlers/asset_handler.go
meletios: We could validate that userID and assetID are correctly formatted, exist in the database, and the user has permission to modify the specified asset

handlers/auth_handler.go
meletios: I chose http.SetCookie to securely store the JWT token in the client's browser, so it is automatically included in subsequent requests, making local testing easier

handlers/asset_handler.go
meletios: Yes, I agree that the UsersFavorites map is coupled with this layer, making it difficult to decouple and test independently.

PR #17, opened 2024-07-09
Go-challenge implementation 

arhsxro:
My implementation for the platform-go-challenge.


api/api_test.go
```diff
@@ -0,0 +1,538 @@
...
+	"github.com/arhsxro/platform-go-challenge/models"
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+)
+
+type MockStore struct {
+	GetUserFavoritesFunc  func(ctx context.Context, userID, filterType string, page, pageSize int) ([]models.Asset, error)
+	AddFavoriteFunc       func(ctx context.Context, userID string, asset models.Asset) error
+	RemoveFavoriteFunc    func(ctx context.Context, userID, assetId string) error
+	UpdateDescriptionFunc func(ctx context.Context, userID, assetID, newDescription string) error
```
nmakro, line 22: Nice way of mocking!

storage/postgres_store.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/arhsxro/platform-go-challenge/models"
+	"github.com/jmoiron/sqlx"
+	_ "github.com/lib/pq"
+)
+
+type PostgresStore struct {
+	db *sqlx.DB
+}
+
+func isValidAssetType(paramType string) bool {
```
nmakro, line 19: Nice that you added a validation. Can you think of a better a way that this could be handled?

storage/postgres_store.go
```diff
@@ -0,0 +1,124 @@
...
+		query = "SELECT asset_id, type, description, data FROM assets WHERE user_id = $1 LIMIT $2 OFFSET $3"
+		err = store.db.SelectContext(ctx, &assets, query, userID, pageSize, offset)
+	}
+
+	return assets, err
+}
+
+// Adds a new favorite asset for a user in the database
+func (store *PostgresStore) AddFavorite(ctx context.Context, userID string, asset models.Asset) error {
+	tx, err := store.db.Begin() //begin transaction
```
nmakro, line 61: Could you describe why you chose to use transactions?

api/handlers.go
```diff
@@ -0,0 +1,240 @@
...
+				errCh <- &models.AssetError{Asset: asset, Err: err}
+			}
+		}(asset)
+	}
+
+	// Wait for all goroutines to finish and close the error channel
+	go func() {
+		wg.Wait()
+		close(errCh)
+	}()
```
nmakro: Can you explain the use of this goroutine?

api/handlers.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Check for errors in goroutines
+	for assetErr := range errCh {
+		if assetErr.Err == context.DeadlineExceeded {
+			log.Println("Request timed out for asset ID ", assetErr.Asset.ID+" and error message: ", assetErr.Err)
+			http.Error(w, "Request timed out", http.StatusGatewayTimeout)
+		} else {
+			log.Println("Error on executing the query for asset ID ", assetErr.Asset.ID+" and error message: ", assetErr.Err)
+			http.Error(w, assetErr.Err.Error(), http.StatusInternalServerError)
```
nmakro, line 178: The information that this error has, what could it be?
Is it safe to send to the outside world these kind of messages?

nmakro:
Hi @arhsxro,

Thank you very much for your time and effort.
I have put some comments in the PR that you can address if you like.
Have a nice day!

storage/postgres_store.go
arhsxro: I could probably move the validation to my handler in order not to proceed more if there is a validation err.
Also I could use a map instead of slice to avoid the loop.

api/api_test.go
arhsxro: Thank you for the feedback!The truth is I spend a lot of time in tests.

storage/postgres_store.go
arhsxro: The truth is that I wasn't 100% sure if i should use a transaction since it only adds/delete/edit  a single asset to the db.Allthough i thought that maybe i get at the same time 2 requests where the first 1 is to delete an asset and the second one is to edit the same asset.So in this case with the use of transaction they do not interfere with each other, maintaining the integrity of the data.Another scenario could be that 1 request is to get the favorite assets and the 2 request to add an asset.

api/handlers.go
arhsxro: Basically in this goroutine i wait for all the goroutines that add an asset to finish in order to close the error channel.By doing so i can then check for any errors in the channel

api/handlers.go
arhsxro: I think that this : assetErr.Err.Error() in the http.Error should not be displayed.It may contains more detailed information about what happened internally.(maybe information for our db or some other sensitive informaiton).

PR #23, opened 2024-10-06
Sven's challenge implementation

svenvermeulen:


loukaspe:
Hey @svenvermeulen, thank you very much for your submission, time and effort for this.

README.md
```diff
@@ -29,3 +29,84 @@ It is appreciated, though not required, if a Dockerfile is included.
...
+        }
+    }
+]'
+
+For full API documentation, visit `http://localhost:8086/swagger/index.html` when the service is running.
+
+```
+
+## About the implementation
+I chose to implement the web server without any data persistence.
```
loukaspe, line 98: If we told you to use a Database system instead of keeping all data in memory
- Which kind of database would you choose and why?
- Can you provide a brief schema of this database (tables definition, relationships, or whatever you would do) ?
- Would you use any ORM to connect and exchange data with the database?

internal/auth/token.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+	"strings"
+
+	"github.com/gin-gonic/gin"
+	"github.com/golang-jwt/jwt/v4"
+	"github.com/google/uuid"
+)
+
+func GetUserIDFromToken(c *gin.Context) (uuid.UUID, error) {
```
loukaspe, line 12: This function is used in every handler you have created. Can you think of every other way you could have used this in the handlers ? Through gin maybe ?

internal/handler/favourites.go
```diff
@@ -0,0 +1,296 @@
...
+	insightRepository *insight.Repository) *FavouritesHandler {
+	return &FavouritesHandler{
+		favouriteRepository: favouriteRepository,
+		audienceRepository:  audienceRepository,
+		chartRepository:     chartRepository,
+		insightRepository:   insightRepository,
+	}
+}
+
+// GetUserFavourites godoc
```
loukaspe, line 40: Nice touch with the Swagger !

internal/handler/favourites.go
```diff
@@ -0,0 +1,296 @@
...
+	"svenvermeulen/platform-go-challenge/internal/repository/favourite"
+	"svenvermeulen/platform-go-challenge/internal/repository/insight"
+	"svenvermeulen/platform-go-challenge/pkg/model"
+
+	"github.com/gin-gonic/gin"
+	"github.com/google/uuid"
+	log "github.com/sirupsen/logrus"
+)
+
+type FavouritesHandler struct {
```
loukaspe, line 21: What are your thoughts on having another layer between Repositories and Handlers ?

internal/handler/favourites.go
```diff
@@ -0,0 +1,296 @@
...
+	favouriteRepository *favourite.Repository
```
loukaspe, line 22: Also, can you think of any advantages on using interfaces to "connect" handler with repository ?

internal/handler/favourites.go
```diff
@@ -0,0 +1,296 @@
...
+// @Router      /favourites [post]
+func (h *FavouritesHandler) CreateUserFavourite(c *gin.Context) {
+	userId, err := auth.GetUserIDFromToken(c)
+	if err != nil {
+		log.Infof("Error obtaining userid from jwt token: %v\n", err)
+		c.AbortWithStatus(http.StatusUnauthorized)
+		return
+	}
+
+	var newFavourite model.UserFavouriteShort
```
loukaspe, line 189: Can you think of any problems of using models of repository layer in the API requests and responses ?
e.g `var newFavourite model.UserFavouriteShort`

internal/handler/favourites.go
```diff
@@ -0,0 +1,296 @@
...
+// @Tags        favourites
+// @Produce     json
+// @Success     200	{object}	[]model.UserFavourite
+// @Failure     400
+// @Failure     401
+// @Failure     404
+// @Failure     500
+// @Router      /favourites/:userid [get]
+func (h *FavouritesHandler) GetFavourites(c *gin.Context) {
+	userId, err := auth.GetUserIDFromToken(c)
```
loukaspe, line 53: I see that c *gin.Context is used here. Could you think of another potential uses of context ?

internal/handler/favourites.go
```diff
@@ -0,0 +1,296 @@
...
+	// Then I wait for all the queries to return and stitch the results back together
+	audienceIDs, insightIDs, chartIDs := h.splitUserFavourites(userFavourites)
+
+	var wg sync.WaitGroup
+	wg.Add(3)
+
+	// AUDIENCES
+	var audiences map[uuid.UUID]*model.Audience
+
+	go func() {
```
loukaspe, line 85: Nice using concurrency and wg !

README.md
svenvermeulen: ### Database type
The kind of database would depend on the non-functional requirements for the project. 
What scalability would it need? Would it be read-heavy or write-heavy? etc.
Having said that, I'd probably default to whatever RDBMS is now being used at the company, which I'd expect to be Postgres, for most services here and only introduce other storage systems if there's a need, and after careful consideration.

### Schema
Let's say favourites, audiences and insights exist in the same DB while charts come from a separate service with its own DB.

#### Main service (favourites-audiences-insights):
```
Audience
- ID uuid (PK)
- Gender: char
- BirthCountry string (FK -> countries.CountryCode)
- AgeFrom int
- AgeTo int
- HoursSpentFrom int
- HoursSpentTo int
- PurchasesFrom int
- PurchasesTo int
* if there were more properties and they don't always have values, they'd probably go into a separate table instead of continuously adding columns.
```
```
Country
- CountryCode varchar (PK)
- Description varchar
```
```
Insight
- ID uuid (PK)
- Description varchar
```
```
User
- ID uuid (PK)
- FirstName varchar
- LastName varchar
- etc etc
```
```
UserFavourite
- ID uuid (PK)
- UserID uuid (FK -> User.Id)
- Description
- ResourceType varchar
- FavouriteId uuid
```
I would probably NOT apply a foreign key constraint on FavouriteId since (a) it's going to refer to different types of entities, some of which (charts) don't exist in this DB and (b) it would potentially affect write performance on this table. 

#### Charts service (its own DB - SQL now but might fit in a NoSQL / document database nicely)

```
Chart
- ID uuid (PK)
- Title string
- XAxisTitle string
- YAxisTitle 
```

```
ChartDataPoint
- ID uuid (PK)
- ChartID uuid (FK -> Chart.ID)
- X float
- Y float
```

Re. the use of an ORM, I'd probably go with whatever is being used at the company, and keep an eye on the actual queries being executed on the RDBMS.

internal/auth/token.go
svenvermeulen: I should've probably added the code to the gin pipeline e.g. `router.use(someJwtCheckFunction)` in `main.go`.
I could abort the request there if a valid token is absent, and store the userId extracted from a valid token in the gin context.

internal/handler/favourites.go
svenvermeulen: In the real world we'd probably also have a domain layer to apply business logic / rules.

internal/handler/favourites.go
svenvermeulen: Using interfaces there would allow us to swap out repository implementations without changing the handlers.
Examples would be separate implementations for different data stores, or injecting a mock implementation for testing purposes.

internal/handler/favourites.go
svenvermeulen: Yes, any changes in the data model would seep into the API, likely as a breaking change.

internal/handler/favourites.go
svenvermeulen: The most obvious example would be binding request/response bodies, accessing parameters and returning HTTP statuses. Other uses are accessing headers, reading client IP.

internal/handler/favourites.go
svenvermeulen: @loukaspe Thanks for reviewing, I added my answers to your comments and questions on the PR. Would be happy to discuss more. Cheers.

loukaspe: Thank you for answering these extra questions. Appreciate it !

internal/auth/token.go
loukaspe: Yes exactly, adding this as a middleware is what I was thinking too.

internal/handler/favourites.go
loukaspe: Yees. By abstracting repository details behind an interface, handlers are not affected if you change how data is stored or accessed, allowing each layer to change independently based on its own concerns. So we ensure that the code can be extended without directly modifying existing components. Also, defining specific, smaller interfaces between handlers and repositories ensures that the handlers only depend on the methods they actually need. All these are examples of how interfaces help us apply the SOLID principles to our code

internal/handler/favourites.go
loukaspe: Generally, passing context.Context from handlers to services and repositories is common practice as it helps us with:
- Cancellation Signals, Timeouts and Deadlines, thus managing the request lifecycle
- Passing Request-Scoped Data, like you said
- Concurrency Management
- Graceful Shutdown of the App

PR #24, opened 2025-01-09
platform-go-challenge solution

mariannapap:
My solution for the platform-go-challenge:
It includes unit tests, docker file and swagger.
For more details please check the ReadMe.
Thank you


lunemec:
Hello @mariannapap ,

Thank you

PR #26, opened 2025-01-23
Ioannis Lampropoulos / GWI Go challenge assignment

lumakos:
This is my implementation on this Go challenge assignment. Thank you for the opportunity to work on this assignment.


lunemec:
Hello @lumakos ,

Thank you

lumakos:
Hi @lunemec,

From my limited experience with Golang, I will try to answer your question by breaking my response into two parts. The first part will cover the current implementation, while the second part will focus on the implementation with an SQL database, such as PostgreSQL. I have mentioned some details about this implementation in the README.md, which certainly changes things a bit compared to the initial implementation.

First of all, this code implements a RESTful API that temporarily stores users' favourite entries using sync.Map. It also includes basic resource management operations (CRUD) and access control through rate limiting. It's following a short analysis of main API endpoint functions.

1. **GetUserFavorites(w http.ResponseWriter, r http.Request)**
    - Retrieves a user's favourite entries based on their user ID.
    - Implements rate limiting to control request frequency.
    - Supports pagination for organising results.
    - Returns an empty list if no entries are found.
   
2. **AddFavorite(w http.ResponseWriter, r http.Request)**
    - Reads the request and decodes an Asset object from the request body.
    - Validates the entry.
    - Generates a description based on the entry type (Chart, Insight, Audience).
    - Assigns a unique ID to the entry.
    - Adds it to the user's list and returns a confirmation response.
    
3. **EditDescription(w http.ResponseWriter, r http.Request)**
    - Updates the description of an existing favorite entry.
    - Verifies that both the user and the entry exist.
    - Returns a success message if the update is successful.

4.  **RemoveFavorite(w http.ResponseWriter, r http.Request)**
    - Deletes a user's favourite entry.
    - Returns an error if the entry is not found.
    - Returns a success message if the entry is successfully removed.

If you run this code for a single user, it will work as expected. Now ,in case of multiple users.

1.  **Concurrent Add Favourite**
If two users add favorites simultaneously:
 - sync.Map.LoadOrStore will create or update the user's entry. But, maybe exists a problem here: When a user adds a new favorite item, LoadOrStore is called first. This means that if there is no existing entry for the userID, an empty list will be created.
 - The function calculateNextId computes the next available ID, which can be critical if many additions happen simultaneously.
- Concern: If two users add favourites at the same time, calculateNextId may experience a race condition if the Store operation doesn't complete in time and maybe generate the same ID. Maybe we need a mutex lock for this issue.

2. **Concurrent Delete Favourite**
If two users try to delete different items:
- Each will have their own list in the sync.Map, so there won't be an issue.
- If they attempt to delete the same item simultaneously, the first request will delete the item, and the second will return "Asset not found."
- Concern: I believe that there is not issue here.

3. **Concurrent Edit Description**
If two users try to edit the same asset, the last request to execute will write the final description.
 - Concern: If two requests change the description simultaneously, the first change may be lost if the data is not reloaded after the initial fetch. The user's asset list is read from UserStore.Load(userID).The specific asset is modified.The new list replaces the old one with UserStore.Store(userID, assets).
 
 On the other side, If you develop a SQL database (for example PostgreSQL) as I mentioned in README.md and use Redis cache for the most frequent queries and deploy it in production with some concurrent users, the application's behaviour will show significant improvements compared to the version that I implemented, as it now uses PostgreSQL for data storage **instead of sync.Map**. Also, it would be nice to have implemented an authentication system using for example JTW, so that the endpoints are not public and only an authenticated user can make requests.
 
1. Every insert/delete operation in the user_favorites, assets, and corresponding data tables must be done via transactions to avoid inconsistencies.
2. The calculateNextId() function is no longer necessary because the database has auto-increment for IDs, so any potential race condition I faced in my initial implementation will be avoided.
3. If there are many concurrent reads and writes (especially in user_favorites), using indexing can help avoid creating locks in the database, preventing performance issues.
4. Using Redis for caching the most frequent requests (e.g., a user's favorites) could reduce the load on the database.
5. The rate-limiting mechanism will work more effectively if Redis is properly configured.

In short, the new system is much more scalable than the initial sync.Map approach.

I am at your disposal for any further questions.

lunemec:
Thank you for your response @lumakos

I have some follow up questions:

> rate limiting to control request frequency

In your **current** implementation, if you remove rate-limiting, what could happen? 

> Every insert/delete operation in the user_favorites, assets, and corresponding data tables must be done via transactions to avoid inconsistencies.

Can you describe what such inconsistency could look like?

> If there are many concurrent reads and writes (especially in user_favorites), using indexing can help avoid creating locks in the database, preventing performance issues.

Can you describe how do you think this locking works without index, and how you think index helps this issue?

> The rate-limiting mechanism will work more effectively if Redis is properly configured.

Can you please describe what do you mean by "properly configured"?

> Supports pagination for organising results.

What other effects does pagination give us? Does it also help in other ways? Would this be different if you use database instead of sync.Map?

lumakos:
Hi @lunemec!

> Thank you for your response @lumakos
> 
> I have some follow up questions:
> 
> > rate limiting to control request frequency
> 
> In your **current** implementation, if you remove rate-limiting, what could happen?
> 

The addition of rate limiting in this case has been implemented for security reasons, given that I have designed the endpoints to be public. For preventing server overload, we use pagination and can also leverage caching. If endpoint was private, I would not use rate limiting.

> > Every insert/delete operation in the user_favorites, assets, and corresponding data tables must be done via transactions to avoid inconsistencies.
> 
> Can you describe what such inconsistency could look like?

For example, we might experience data loss, data insertion might fail, leading to inconsistency, or there could be a reference to an asset that lacks complete data.For this reason, we use transactions, which follow a sequence of SQL commands. This means that either all commands execute successfully (and are committed to the database), or none of them execute at all (in case of failure, a rollback occurs). If any of the SQL commands fail, all changes are canceled (ROLLBACK), preventing inconsistency.(ACID properties in DBMS)
 
> > If there are many concurrent reads and writes (especially in user_favorites), using indexing can help avoid creating locks in the database, preventing performance issues.
> 
> Can you describe how do you think this locking works without index, and how you think index helps this issue?

 Indexes can help reduce the number of locks required during operations by allowing the database engine to quickly locate the necessary rows, thereby reducing the scope of the locks needed. For example, without an index, an update operation might require scanning the entire table and locking every row, which can lead to blocking and deadlocks.However, with an appropriate index, the database can lock only the specific rows that need to be updated, minimizing the impact on other transactions. This means that other users attempting to read or write data may be blocked until the first transaction completes. This is where indexing helps by managing locks and speeding up queries. For example, with an index, the database can quickly locate relevant records and lock only the necessary rows instead of the entire table. Also, an index on user_id and asset_id allows the database to access records without performing a full table scan and run the corresponding update operation.
 
> > The rate-limiting mechanism will work more effectively if Redis is properly configured.
> 
> Can you please describe what do you mean by "properly configured"?

We choose Redis with sufficient RAM so that it can store the rate limiting counters in memory, avoiding the risk of the memory filling up and forcing Redis to start evicting data or crashing.Next, proper expiration management is essential. User counters should be automatically deleted when the rate limit time window expires. If this is not handled correctly, they will take up memory unnecessarily.Finally, if we have multiple Redis nodes, inconsistencies may arise due to network delays. So its better to  use a Redis Cluster for better scaling and to avoid a single point of failure.

> > Supports pagination for organising results.
> 
> What other effects does pagination give us? Does it also help in other ways? Would this be different if you use database instead of sync.Map?

Instead of returning all of a user's favorites every time they request the data, pagination breaks the results into smaller chunks (e.g., 10 items at a time).Additionally, with pagination, each request retrieves only the necessary data, preventing large responses that could cause issues such as overloading database. With sync.Map, however, all the data is first loaded into memory and then manually paginated. This is inefficient if there are many users with large favorite lists.

PR #27, opened 2025-02-09
Platform-go-challenge Panagiotis

ghost:
Hello 👋, this is my solution for the platform-go-challenge


lunemec:
Hello @PanaManta ,

Thank you

test/controllers/favorites_controller_test.go
```diff
@@ -0,0 +1,128 @@
+package tests
```
Beny406, line 1: Hi, thx for submitting. Can you think of a way how to improve a test coverage without adding additional tests?

ghost:
> Hello @PanaManta , thank you for your submission.
> 
> My first question on your submission is, what will happen if I take the code, and deploy it to production, and have few concurrent users using it? Can you describe what will happen, in as much detail as you can think of?
> 
> Thank you

Hi @lunemec, thx for reviewing.

Sure I will try to describe what each individual layer will do. In the end I will outline some scaling considerations and propose some improvements

# Flow description

## 1. HTTP request handling + controller layer
The Gin framework will spawn a separate goroutine for each request allowing the requests to be handled independently.

### 1.1 Routing 
Each individual request will be routed to the registered endpoint based on the HTTP method and URL path.

### 1.2 Authentication middleware
Before reaching the controller each request will pass through the JWT middleware for authentication, which:

- extacts the `JWT` token from the request headers
- validates the `token` and verifies the user identity
- attaches the `user_id` to the request context

If authentication fails the request is immediatly rejected without any specific error for security purposes.

### 1.3 Favorites controller
Once authenticated, the request is forwarded to the Favorites Controller, which:

- extracts request parameters (e.g., `user_id`, `asset_id`)
- calls the Favorites Service to perform the business logic
- formats the response and sends it back to the client

## 2. Service layer
The favorites service acts as an abstraction over the repository, which:

- ensures bussiness logic is correctly applied
- calls the repository to fetch, add, deletet  over our store (in memory)
- returns the result of the opration to the controller

## 3. Repository layer
The repository is responsible for data storage. We are currently using an in-memory map to store user favorites

- locks the store, with the introduced mutex. (**bottleneck**)
- performs the operation
- returns the result to the service layer

# Scaling considerations
The **1, 2 (controller + service layer)** can be easily horizontally scaled and with the help of a load balancer and a rate limiter, can provide efficient high concurrency

However on the **3 (repository layer)** this is not true because of the following reasons:

- as more users try to `add/delete` favorites, they will be queued behind the lock.
- if the app is restarted, **all data is lost**.
- if millions of users interact with the app, the in-memory map grows uncontrollably.
- cannot be horizontally scalled since the storage lives in memory and cannot be shared between instances
- currently no **pagination** exists

# Improvements

1. Migrate to **sharded** locks that are defined based on the user_id. For example we could have a store with 10 locks and each user could use the `user_id % 10` lock.

   This could have the following benefits:

   - each shard has its own lock, allowing parallel writes across different shards.
   - reduce contention and improve performance

2. Introduce a persistend storage like Postgres

   - improves horizontal scaling since the data are not in memory and can be shared among instances
   - builtin locking mechanisms
   - prevents data loss on restarts

3. Add pagination + caching

   - fetching a subset of favorites for each user, instead of fetching all documents at the same time, reducing memory consumption and speeding up response times as well as network utilization
   - cache frequenlty fetched favorites for improving performance and response time and limiting round trips to the DB

5. Security

   - add rate limiting to prevent abuse and ddos attack
   - prevent timing attacks in the JWT authentication middleware by ensuring that all error responses take the same amount of time before returning

test/controllers/favorites_controller_test.go
ghost: Hi @Beny406, thx for reviewing. Yes one way would be to remove the mocking of the service layer and let the actual implementation execute as much as possible.

However, to ensure test stability and avoid flakiness, we would need to mock the non-deterministic functions, such as `math/rand` which is used to determine the asset to populate at runtime, as well as filling the chart points.

PR #28, opened 2025-02-16
GWI Engineering Challenge Submission – Favorites API

georgelos825:
Submitting the GWI Engineering Challenge project – a Go-based Favorites API with full CRUD functionality, JWT authentication, HTTPS support, concurrency handling, Dockerization, and robust error handling. The provided README contains detailed instructions and implementation notes. Thank you for the opportunity to showcase my skills!


lunemec:
Hello @georgelos825 ,

Thank you

georgelos825:
Hello @lunemec. Thank you for the question. I’d be happy to describe how the system would behave under production conditions with concurrent users:

1. Concurrency in Favorites Retrieval: 
The GetFavoritesHandler is designed to retrieve different asset types concurrently using sync.WaitGroup, ensuring faster responses by parallelizing the retrieval of charts, insights, and audiences.
Internal synchronization is handled using sync.Mutex to prevent race conditions when appending to the shared slice of results.

2. Concurrency in Asset Storage:
The in-memory storage uses sync.Mutex for read and write operations, ensuring data consistency.
Since this is an in-memory map, read operations are very fast, but high write concurrency may lead to contention due to the global mutex lock.

3. Handling Multiple Users:
Each user has a dedicated map entry for their assets, which reduces contention since different users don’t block each other.
However, if many concurrent requests from the same user occur, they will be serialized due to the mutex protecting that user’s map entry.

4. JWT Token Validation:
The JWT middleware is lightweight and efficiently validates tokens with minimal performance impact.
Token storage ensures only one valid token per user, which prevents abuse without adding overhead to standard requests.

5. HTTPS Overhead:
HTTPS is enabled using self-signed certificates for local and Docker deployments. In production, replacing these with valid SSL certificates would maintain security with minimal performance overhead.

6. Potential Bottlenecks Under High Load:
The primary bottleneck would be write operations due to mutex locking during add, edit, and delete actions.
Concurrent read operations are efficient but still require locking due to the map’s nature.
For large-scale production usage, using a concurrent-safe data structure like sync.Map or transitioning to an SQL database would improve scalability.

7. Behavior Under Heavy Load:
With a moderate number of concurrent users, the application should remain responsive due to the efficient Go runtime and lightweight Gin framework.
As concurrency increases, the lock contention on shared maps could lead to slower writes, though reads should still be relatively fast.

In summary, the application should perform well under typical concurrent usage, especially for read-heavy workloads. For heavy concurrent writes, optimizing the locking mechanism or transitioning to a production-grade database would improve performance and scalability.

georgelos825:
@lunemec kind reminder.

Στάλθηκε από Outlook για Android<https://aka.ms/AAb9ysg>
________________________________

PR #32, opened 2025-03-18
George Semertzakis

gsemer:


cmd/main.go
```diff
@@ -0,0 +1,95 @@
...
+		return value
+	}
+	return fallback
+}
+
+func main() {
+	log.Println("Server is up and running...")
+
+	log.Println("Waiting 10 seconds for the ArangoDB container to be ready...")
+	time.Sleep(10 * time.Second)
```
lkslts64, line 31: For the purposes of this assignemnt, this is fine. However, in a production setting how would you approach the same problem (i.e. one service depends on another)? Is the current implementation always reliable?

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: '3.7'
+
+networks:
```
alexandros-georgantas, line 3: could you please explain why this is needed?

docker-compose.yml
```diff
@@ -0,0 +1,41 @@
...
+  platform-network:
+    external: true
+
+services:
+  arangodb:
+    image: arangodb:latest
```
alexandros-georgantas, line 9: could you elaborate on the use of the `latest` tag? In a production environment is this considered a good approach?

domain/assets.go
```diff
@@ -0,0 +1,46 @@
+package domain
+
+import "time"
+
+type Asset struct {
+	Key         string    `json:"_key,omitempty"`
```
alexandros-georgantas, line 6: The decision of using this `Key` attribute I assume is idiomatic to ArangoDB? Could you provide an SQL alternative for all the models you have implemented? Also, discuss a bit any decisions you will make on potential indexes.

domain/favourites.go
```diff
@@ -0,0 +1,23 @@
+package domain
+
+import "time"
+
+type Favourite struct {
+	From      string    `json:"_from,omitempty"` // Reference to the user document
```
alexandros-georgantas, line 6: I can also see that this is idiomatic to ArangoDB. Could you please elaborate a bit on your decision to use this database solution?

persistence/assets.go
```diff
@@ -0,0 +1,51 @@
...
+
+type AssetRepository struct {
+	collection driver.Collection
+}
+
+func NewAssetRepository(collection driver.Collection) *AssetRepository {
+	return &AssetRepository{collection: collection}
+}
+
+func (ar AssetRepository) CreateAssets() ([]domain.Asset, error) {
```
alexandros-georgantas, line 21: Could you briefly mention how would you handle seeding data as well as migrating data in a production environment?

persistence/assets.go
```diff
@@ -0,0 +1,51 @@
...
+			log.Printf("Couldn't create new document")
+			return []domain.Asset{}, err
+		}
+		asset.Key = meta.Key
+		arangoAssets = append(arangoAssets, asset)
+	}
+	return arangoAssets, nil
+}
+
+func (ar AssetRepository) GetAsset(assetID string) (string, error) {
```
alexandros-georgantas, line 45: I would have expected by the name of this method to have a different return type. What is your rationale behind that? Would you consider refactoring this in the future and how (in the entire context of your source code)?

persistence/favourites.go
```diff
@@ -0,0 +1,134 @@
...
+			Description: asset.Description,
+			CreatedAt:   asset.CreatedAt,
+			Data:        asset.Data,
+		}
+		assets = append(assets, asset)
+	}
+	return assets, nil
+}
+
+func (fr FavouriteRepository) EditFavourites(userID, assetID, description string) (domain.Asset, error) {
```
alexandros-georgantas, line 71: Could you explain what will happen when two users try to edit an asset's description concurrently?

presentation/favourites.go
```diff
@@ -0,0 +1,129 @@
...
+		writer.Write(result)
+		return
+	}
+
+	writer.WriteHeader(http.StatusOK)
+	result, _ := json.Marshal(asset)
+	writer.Write(result)
+}
+
+func (fh FavouriteHandler) DeleteFavourite(writer http.ResponseWriter, request *http.Request) {
```
alexandros-georgantas, line 105: Would you consider changing the response for something more RESTful?

presentation/routes.go
```diff
@@ -0,0 +1,26 @@
+package presentation
+
+import "platform-go-challenge/domain"
+
+func CreateRoutes(fs domain.FavouriteService) map[string]domain.RouteDefinition {
+	fh := NewFavouriteHandler(fs)
+
+	return map[string]domain.RouteDefinition{
```
alexandros-georgantas, line 8: Will you consider changing these endpoints to something more RESTful?

alexandros-georgantas (commented):
Hello @gsemer!

Thanks for taking the time to complete the assignment!
Overall, you did a good job — well done!

I've left some comments for you as part of the PR review, so please feel free to proceed with addressing them.

cmd/main.go
```diff
@@ -0,0 +1,95 @@
...
+	favouriteService := services.NewFavouriteService(favouriteRepo, userRepo, assetRepo)
+
+	favouriteRoutes := presentation.CreateRoutes(favouriteService)
+	for routePath, routeDefinition := range favouriteRoutes {
+		log.Printf("adding %s route with methods %v\n", routePath, routeDefinition.Methods)
+		r.Handle(routePath, routeDefinition.HandlerFunc).Methods(routeDefinition.Methods...)
+	}
+
+	http.Handle("/", r)
+	log.Fatal(http.ListenAndServe(":"+serverOnPort, r))
```
lkslts64: Can you please explain how would you change the code so that your HTTP server shuts down gracefully (no need to change the code)? When your application is deployed on Kubernetes, why it is important to gracefully shutdown an HTTP server?

cmd/main.go
```diff
@@ -0,0 +1,95 @@
...
+
+func main() {
+	log.Println("Server is up and running...")
+
+	log.Println("Waiting 10 seconds for the ArangoDB container to be ready...")
+	time.Sleep(10 * time.Second)
+
+	arangoURL := getEnv("ARANGO_URL", "")
+	arangoUsername := getEnv("ARANGO_USERNAME", "root")
+	arangoPassword := getEnv("ARANGO_PASSWORD", "")
```
lkslts64, line 35: In a real application, how would you manage the app's secrets?

presentation/favourites_test.go
```diff
@@ -0,0 +1,130 @@
...
+	"platform-go-challenge/consts"
+	"platform-go-challenge/domain"
+	"reflect"
+	"testing"
+	"time"
+
+	"github.com/gorilla/mux"
+)
+
+func TestAddToFavourites(t *testing.T) {
```
lkslts64, line 19: Nice test coverage!

persistence/favourites.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	return favourite, nil
+}
+
+func (fr FavouriteRepository) GetFavourites(userID string) ([]domain.Asset, error) {
+	bindVars := map[string]interface{}{
+		"userID": userID,
+	}
+
+	query := `FOR f IN favourite
```
lkslts64: As the number of user's favourites grows, this query becomes slower. In cases where a user possess a lot of favourites, this query could become a bottleneck. 

What can you do to guarantee that the query always executes within a short/predictable timespan, regardless of the user's favourites?

lkslts64 (commented):
Hey @gsemer, thanks for completing the assignment, we appreciate your effort! 

Overall, you did very well. You followed Domain-Driven Design (DDD) principles, you added an generous suite of tests, support for Docker etc. Good job!

I also left some comments for you to check.

Thanks again.

docker-compose.yml
gsemer: For the purpose of the project, I could avoid using it.
In general, it is needed, because it provides more control and flexibility.
Also, I would say for security reasons, i.e. if two teams use the same network, then one team has access to the other team's project, data etc.
Furthermore, for better communication between different containers. They can communicate by name.

docker-compose.yml
gsemer: It is not recommended to use the latest tag in a production environment.
First of all, updates on the image might break the app without even changing the code.
Furthermore, the dev, staging and production environments might use different versions of the image, which will make debugging and testing unreliable.
Last, but, of course, not least, it can be a security risk.

persistence/favourites.go
gsemer: Unfortunately, I didn't have the time to make the code work concurrently. So, the last write wins! That means that there will be no warnings, no errors and probably data loss.
However, the same would happen even if the code was running concurrently.
What I would do? I would try to find a way to raise a conflict error in case two users try to edit a document at the same time.

cmd/main.go
gsemer: Honestly, I would never use this, even in a dev environment, meaning that the current implementation is not always reliable!
If you take a look in docker compose file, I made the app in a such a way so that it depends on the container in which ArangoDB is running. 
However, for some reason, it didn't work and I didn't have much time to search what causes this behaviour.

cmd/main.go
gsemer: In a real application, there are two ways to manage the secrets:
1. Environment variables, and
2. Secret management services (Kubernetes or portainer have such a system to store secrets). 

Of course, there are more.

persistence/assets.go
gsemer: The thing is that in ArangoDB every element of a collection has an id and a key. 
The difference between those two is that id={name of the collection}/key.
For some reason, I forgot to add the ID in the domain.
Of course, I would consider refactoring this in the future.
One possible solution would be to include the ID in the domain and then the function would return the whole entry. In such a way, I would be able to use the ID in the next steps.

presentation/routes.go
gsemer: Yes, I would consider it, because it improves clarity!

Following the approach of the above comment, I would change the code into
```
func CreateRoutes(fs domain.FavouriteService) map[string]domain.RouteDefinition {
	fh := NewFavouriteHandler(fs)

	return map[string]domain.RouteDefinition{
		"/": {
			Methods:     []string{"POST"},
			HandlerFunc: fh.AddToFavourites,
		},
		"/": {
			Methods:     []string{"GET"},
			HandlerFunc: fh.GetFavourites,
		},
		"/{asset_id}": {
			Methods:     []string{"PUT"},
			HandlerFunc: fh.EditFavourites,
		},
		"/{asset_id}": {
			Methods:     []string{"DELETE"},
			HandlerFunc: fh.DeleteFavourite,
		},
	}
}
```
and, of course, modify the rest of the code accordingly.

cmd/main.go
gsemer: I would do the following: 
1. Create a server that will run in a go routine. The advantage is that my main function can listen for shutdown signals in parallel.
2. Listen for OS signals (SIGTERM).
3. Use http.Server.Shutdown(context) to give my server time to finish the ongoing requests.

The application will run in a pod in Kubernetes. So, if that pod is terminated and the default grace period exceeds the limit, then Kubernetes force shutdown to the application. That could leave HTTP requests unfinished or it could interrupt database editing etc.

persistence/assets.go
gsemer: Firstly, I would store only necessary data avoiding potential duplications. For example, if I had used SQL, I could have the following table:
```
type AssetType struct {
    ID   uuid.UUID `gorm:"type:uuid;primaryKey" json:"id"`
    Name string    `gorm:"uniqueIndex" json:"name"`
}
```
Those data would be either stored or be ignored once the server is up, depending on their existence or not. Also, I would distinguish between staging and production data. I could achieve that with an environment variable.

Secondly, regarding data migration in a production environment and before proceeding with that, I would definitely make sure that it works first on dev and staging. Also, I would always create a backup of the database before proceeding with any migration. The reason is that, if a problem occurs after migration, I need to be ready for a rollback, which will be of high(est) priority!

presentation/favourites.go
gsemer: Yes, of course! 
First of all, I would create a subrouter in main file `favouritesRouter := router.PathPrefix("/favourites").SubRouter()`. This approach will help us to distinguish between different subrouters in case the project grows.
Secondly, I would handle the errors by using `http.Error(w http.ResponseWriter, error string, code int)`.
Finally, I would only return as a response only `http.StatusOK` in case the favourite is removed from user's favourites with an empty body.

PS: The first modification is not connected with this question, it will just make the code cleaner. However it is connected with the following question, the one about making the endpoints more RESTful.

domain/assets.go
gsemer: For this project, I would add an index to `UserID` in Favourite structure. Using an index on this field allows to get quickly all the favourite assets of a user. Outside of the scope of this project, if, for example, we've been asked to get all the assets of a particular type, then I would add an index to `AssetType` to speed up the search and to avoid scanning all the rows of the table.

Exactly, the `Key` attribute plays the role of a primary key.

Below I provide an SQL alternative for all the models:
```
type User struct {
    ID    uuid.UUID    `gorm:"primaryKey" json:"id"`
    FirstName    string    `json:"first_name"`
    LastName    string    `json:"last_name"`
    Username    string    `gorm:"uniqueIndex" json:"username"`
    Email    string    `gorm:"uniqueIndex" json:"email"`
    CreatedAt    time.Time    `json:"created_at"`

    Favourites    []Favourite    `json:"favourites" gorm:"foreignKey:UserID"`
}

type Asset struct {
    ID    uuid.UUID    `gorm:"primaryKey" json:"id"`
    Type    string    `json:"asset_type"`
    Description    string    `json:"description"`
    CreatedAt    time.Time    `json:"created_at"`
    Data    AssetData    `json:"data"`
    
    Favourites []Favourite    `json:"favourites" gorm:"foreignKey:AssetID"`
}

type Favourite struct {
    ID    uuid.UUID    `gorm:"primaryKey" json:"id"`
    UserID    uuid.UUID    `gorm:"index" json:"user_id"`
    AssetID    uuid.UUID    `gorm:"index" json:"asset_id"`
    CreatedAt    time.Time    `json:"created_at"`
}

(plus the rest structures, which will NOT be models in database)
```

persistence/favourites.go
gsemer: Honestly, I would just change the query, because it is indeed terrible! Assuming that F and A denotes the number of favourites and the number of assets in database respectively, then, since I didn't use any potential indices, the time complexity equals O(F*A)!

Instead, I would use the following query:
```
FOR v, e IN 1..1 OUTBOUND @userID favourite RETURN v
```

How it works?
It follows the process of a graph traversal.
A traversal starts at a specific document (start vertex) and follows all edges connected to it. For all vertices that are targeted by there edges, it will again follow all edges connected to them and so on. It is possible to define how many of these follow iterations should be executed at least (min depth) and at most (max depth).
In this case, it starts at the user vertex, hops 1 step using the 'favourite' edge and visits all user's favourite assets.
The time complexity is O(N), where N denotes the out-degree of the starting vertex.

Below, I share the documentation:
[traversals-explained](https://docs.arangodb.com/stable/aql/graphs/traversals-explained/)

domain/favourites.go
gsemer: 1. Queries
    Quering database becomes simpler in this case. SQL can become complex with multiple JOINs, whereas AQL becomes easier to understand, since it is a 1-hop graph traversal.
2. Schema flexibility 
    Using graph databases makes it easier to handle different asset types without multiple tables.

PR #33, opened 2025-05-30
implement user favorite management features

antonioualex:
# GWI Favorites Asset Service - Implementation

## Overview
This PR implements a user favorites management system for the GWI platform, allowing users to manage personalized lists of favorite assets (Charts, Insights, and Audiences) through a RESTful API.

### Core API Endpoints
- `GET /favorite/` - Retrieve all user favorites
- `POST /favorite/{aid}` - Add asset to favorites
- `PATCH /favorite/{id}` - Update favorite description
- `DELETE /favorite/{id}` - Remove from favorites

### Database & Persistence
- **PostgreSQL** with Docker containerization
- **Migration System** for schema management and data seeding
- **Repository Pattern** for data access abstraction

### Technology Stack
- **Go 1.24+** with modern best practices
- **PostgreSQL** for data persistence
- **Docker & Docker Compose** for containerization
- **Testcontainers** for integration testing
- **Mockgen** for test mocking
- **Bun** ORM for PostgreSQL

### Testing Strategy
- **Unit Tests**: Individual component testing with mocks
- **Integration Tests**: Database integration with Testcontainers
- **Test Coverage**: Comprehensive coverage across all layers
- **CI/CD**: GitHub Actions for automated testing

## 📁 Project Structure
```
├── app/                    # Application layer (use cases, business logic)
├── cmd/                    # Application entry points
├── domain/                 # Core domain entities and interfaces
├── infra/                  # Infrastructure setup (Docker, migrations)
├── internal/               # Internal packages (config, server setup)
├── persistence/            # Data access layer (repositories)
└── presentation/           # HTTP handlers and routing
```


cmd/main.go
```diff
@@ -0,0 +1,31 @@
...
+		slog.Error("failed to run the application", "error", err.Error())
+		cleanExit(1)
+	}
+}
+
+// cleanExit is used to exit the application while ensuring deferred functions are executed
+func cleanExit(code int) {
+	// Allow deferred functions to run before exiting
+	defer os.Exit(code)
+}
```
alesr, line 31: Could you please elaborate on what is happening here?

The server returns an error, logs the error and then?

internal/server/server.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return runServer(ctx, httpServer, cfg.Server.ShutdownTimeout)
+}
+
+// setupDatabase initializes the database connection using the provided configuration.
+func setupDatabase(cfg config.DatabaseConfig) (*bun.DB, error) {
+	sqldb := sql.OpenDB(pgdriver.NewConnector(
+		pgdriver.WithNetwork(cfg.Network),
+		pgdriver.WithAddr(cfg.Address),
+		pgdriver.WithTLSConfig(nil),
```
alesr, line 47: Why do we need to pass `nil` to the `WithTLSConfig`?

Is TLS enabled by default? What does it mean to add this flag and pass a nil configuration?

internal/server/server.go
```diff
@@ -0,0 +1,137 @@
...
+	"pgc/persistence/postgres"
+	"pgc/presentation"
+)
+
+// Run initializes the GWI application, sets up the database, services, and HTTP server,
+// and starts the server with graceful shutdown capabilities.
+func Run(ctx context.Context, cfg *config.Config) error {
+	slog.Info("Starting GWI application...")
+
+	db, err := setupDatabase(cfg.Database)
```
alesr, line 26: What would be the benefit of not initializing the database here, but pass it as an argument to the run function?

app/favorite.go
```diff
@@ -0,0 +1,64 @@
...
+	switch {
+	case favorite.UserID <= 0:
+		return domain.Favorite{}, domain.ErrInvalidUserID
+	case favorite.AssetID <= 0:
+		return domain.Favorite{}, domain.ErrInvalidAssetID
+	}
+
+	favorite.ID = 0 // Ensure ID is zero (default value) for insertion
+
+	return s.favoriteRepository.Insert(ctx, favorite)
```
alesr, line 34: What happens to the insert if the user closes the connection with the server?

alesr (commented):
Hi @antonioualex, thanks so much for submitting your solution for the assignment! I’ve left a couple of questions for you to take a look at when you have a moment. No rush, take your time. 😊

cmd/main.go
antonioualex: The server returns an error, logs the error and then calls cleanExit(1). Inside cleanExit(), os.Exit(1) is scheduled as a deferred function. When cleanExit() finishes, the deferred os.Exit(1) executes immediately. The cleanExit function doesn't actually accomplish what its comment and name suggests creating confusion about when exit will occur. Besides that, there's nothing left in that scope to run. I would remove cleanExit() entirely and use os.Exit(1) directly in main() so it can be clearer and doesn't create false expectations.

internal/server/server.go
antonioualex: It means no custom TLS configuration is provided which overrides the default behavior. When I pass `nil` to `WithTLSConfig()`, the connector will use disabled ssl mode which is needed for local development. The correct approach would be to use, pgdriver.WithInsecure(cfg bool) and pass the value through configuration.

internal/server/server.go
antonioualex: 🤔 It would give better testability by giving the opportunity to inject a mock database.

app/favorite.go
antonioualex: A context cancellation will occur and if it is before or during the database operation execution, then it will not be performed and an error `context.Canceled` will be returned. In this case it will also log user info and will return an error InternalServerError in the application layer, which is not correct, should return (and handle) a different type of error. If context cancellation occurs after the database write completes, then the data will be stored in the database and the connection will close before the success response reaches the client.

antonioualex:
Hi @alesr, thank you very much for your review, appreciate it!! 😊 😄 😄

PR #34, opened 2025-05-30
GWI Favorites API

gitvam:
Full solution submitted by George Vamvakousis.

Thanks for the opportunity!


internal/store/postgres_store.go
```diff
@@ -0,0 +1,205 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := db.Ping(); err != nil {
+		return nil, err
+	}
+	return &PostgresStore{db: db}, nil
+}
+
+func (ps *PostgresStore) ListFavorites(userID string, limit, offset int) ([]models.Asset, error){
```
pamanta, line 32: Can you identify any potential issues with this method in terms of performance and pagination and what would you do to address them?

cmd/server/main.go
```diff
@@ -0,0 +1,56 @@
...
+			sr.Post("/", h.AddFavorite)
+			sr.Delete("/{assetID}", h.RemoveFavorite)
+			sr.Patch("/{assetID}", h.EditFavoriteDescription)
+		})
+	})
+
+	log.Println("Server running on http://localhost:8080 ...")
+	if err := http.ListenAndServe(":8080", r); err != nil {
+		log.Fatalf("could not start server: %v", err)
+	}
```
pamanta, line 55: What happens when the server receives a signal to shutdown (ex. kubernetes environment)? How would you approach a solution for a graceful shutdown flow?

internal/store/postgres_store.go
```diff
@@ -0,0 +1,205 @@
+package store
+
+import (
+	"database/sql"
```
pamanta, line 4: Could you walk me through your decision process of choosing raw sql, did you consider other alternatives?

README.md
```diff
@@ -1,31 +1,258 @@
...
-Of course, usually we would choose to solve any given requirement with the simplest possible solution, however that is not the spirit of this challenge.
+All endpoints require a valid JWT token in the `Authorization: Bearer <token>` header.
 
-## Challenge
+| Method | Path                                              | Description                             |
+|--------|---------------------------------------------------|-----------------------------------------|
+| GET    | `/v1/users/{userID}/favorites`                    | List all favorite assets for the user   |
+| POST   | `/v1/users/{userID}/favorites`                    | Add a new favorite asset                |
+| DELETE | `/v1/users/{userID}/favorites/{assetID}?type=...` | Remove a favorite by external ID & type |
+| PATCH  | `/v1/users/{userID}/favorites/{assetID}?type=...` | Edit description of a favorite asset    |
```
pamanta, line 18: Why is `type` required to access a specific favourite? Would you consider redesigning these 2 endpoints to something a bit more restful and how would you do it?

pamanta (commented):
Hi @gitvam, thank you very much for submitting your solution to the assessment. I've left some questions for you to take a look when you have the time, no rush :)

internal/store/postgres_store.go
gitvam: Yep, the current implementation runs one query per asset type, each with its own pagination. That means limit/offset aren't applied globally, and we could return more items than expected or miss consistency between pages. I’d unify the queries with a UNION ALL and paginate after the merge to ensure consistent ordering and better performance.

cmd/server/main.go
gitvam: Good catch! At the moment, if the server receives a shutdown signal (like SIGTERM), it just exits immediately, any in-flight requests are dropped, and there’s no cleanup.
To handle this properly, I’d set up signal handling, run the server in a goroutine, and use http.Server.Shutdown with a timeout. That way, the server can stop accepting new requests, finish what it’s doing, and shut down cleanly.

internal/store/postgres_store.go
gitvam: I considered using GORM but kept things simple with raw SQL since the scope was small. That said, using an ORM here definitely could’ve helped reduce some of the boilerplate, in a larger project I’d go that way for sure.

README.md
gitvam: Yeah, fair point. I used asset_type and external_id to identify favorites, which is why the type param is there on both DELETE and PATCH.
I could’ve just exposed the favorite’s unique ID and treated it as its own resource, like DELETE /favorites/{id} and PATCH /favorites/{id}, which would’ve been more RESTful and cleaner overall.

gitvam:
@pmantafounis-gwi thanks for the thoughtful review, appreciate it! I’ve gone through all the comments and added replies. Let me know if there’s anything else you’d like me to clarify.

PR #40, opened 2025-09-11
GlobalWebIndex Engineering Challenge: Submission 

burrito-eng:


data/dbData.go
```diff
@@ -0,0 +1,156 @@
+package data
+
+type Users struct {
+	UserId int    `json:"UserId"`
+	Name   string `json:"name"`
+}
+
+var User = []Users{
+	{UserId: 1, Name: "John Doe"},
```
alexandros-georgantas, line 9: The naming convention is a bit weird here. The context is User, could you elaborate about your decision for repeating that in your attribute e.g. UserId?

data/dbData.go
```diff
@@ -0,0 +1,156 @@
...
+	{UserId: 2, Name: "Jane Doe"},
+	{UserId: 3, Name: "Alan Doe"},
+	{UserId: 4, Name: "Bob Doe"},
+	{UserId: 5, Name: "Alice Doe"},
+}
+
+// All assets
+
+type AllAssets struct {
```
alexandros-georgantas, line 18: Is the All prefix of `AllAssets` necessary here, what is its value? Also, could you suggest any alternatives instead of having `AllAssets` as struct what other type could work here?

endpoints/AddAsset.go
```diff
@@ -0,0 +1,64 @@
...
+
+func AddAsset(w http.ResponseWriter, r *http.Request) {
+
+	var req AddAssetRequest
+	w.Header().Set("Content-Type", "application/json")
+	err := json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		resp := AddAssetResponse{
+			Message: "Invalid JSON",
+			Code:    200,
```
alexandros-georgantas, line 30: Could you elaborate on your decision of using this status code? Could you suggest any alternatives?

endpoints/DeleteAsset.go
```diff
@@ -0,0 +1,63 @@
...
+	for i := len(data.FavAsset) - 1; i >= 0; i-- {
+		if data.FavAsset[i].AssetId == req.AssetId && data.FavAsset[i].UserId == req.UserId {
+			data.FavAsset = append(data.FavAsset[:i], data.FavAsset[i+1:]...)
+			success_deletion = true
+			break
+		}
+	}
+
+	fmt.Println("Favorite Assets after deletion are :", data.FavAsset)
+	resp := DeleteAssetResponse{
```
alexandros-georgantas, line 48: Could you please explain why you've decided to return this response for the `DELETE` request? Could you suggest an alternative for both the resp and the status code?

endpoints/EditDescriptionAsset.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+		json.NewEncoder(w).Encode(resp)
+		return
+	}
+
+	var success_change bool = false
+
+	for i := range data.Assets {
+		if data.Assets[i].ID == req.AssetId {
+			data.Assets[i].Description = req.NewDescription
```
alexandros-georgantas, line 41: Could you discuss if this operation here is safe on an environment with multiple users? If you had time how would you handle concurrency?

main.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"net/http"
+	"os"
+	"platform-go-challenge/endpoints"
+)
+
+func main() {
```
alexandros-georgantas, line 11: Could you explain why you chose to have a different endpoint for these actions? Is it this a RESTful approach? How would you refactor that?

alexandros-georgantas (commented):
Hi @burrito-eng 👋 
Thank you for spending time working on this assignment!
I have left some comments for your PR and I am looking forward checking your answers!
Thanks,
Alex

data/dbData.go
burrito-eng: I modeled the structs as if they were database tables, which is why I used names like `UserId` to act as a primary key. Maybe an `id` it will not be so confusing and more Go friendly.

endpoints/AddAsset.go
burrito-eng: I initially used `200` to ensure the frontend always receives a JSON body for displaying messages to make better the user experience. However, I realize it’s more accurate to return `400 Bad Request` for invalid `JSON` while keeping the `JSON` body with a message.

endpoints/DeleteAsset.go
burrito-eng: I initially returned a `success: true/false` and a `message` field in the JSON response so the frontend could display user-friendly messages for both success and failure cases. However i can return the status code `204 No Content` and not message in an successful delete or `404 Not found` if the `id` does not exists with an appropriate JSON message.

data/dbData.go
burrito-eng: You’re right, the `All` prefix doesn’t add much value here. A clearer name would simply be `Asset`. My thinking was to model this as if it were a database table holding all the assets, which is why I called it `AllAssets`. Maybe after some research i can use an `interface` that can combine different types in the same slice.

endpoints/EditDescriptionAsset.go
burrito-eng: This operation it's not safe because multiple users could modify the `struck` at the same time without an `mutex`. If i have more time i would use a database, which would make `UPDATE` operation both more accurate and safe under concurrent access.

main.go
burrito-eng: I originally used separate endpoints for adding, editing, and deleting favorites because each action required slightly different data in the request body and had different processing logic. For example, adding a favorite needs `user_id` and `asset_id`, editing a description needs `asset_id` and `new_description`, deleting a favorite needs `user_id` and `asset_id` and fetching the data need the `asset_id`. While these endpoints share some common elements, I found it simpler for this exercise to separate them. I understand that in a RESTful design, it would be more standard to use a single URL `(/users/{id}/favorites)` for actions on associated information, and different actions using `HTTP` methods `(GET, POST, PUT, DELETE)` instead of always using `POST`. This would make the API more easier to maintain in a larger project.

burrito-eng:
Hello @alexandros-georgantas, thanks for reviewing my assignment. I have provided answers to all the review comments.
Please let me know if any further clarification is needed.

PR #41, opened 2025-09-11
GlobalWebIndex Engineering Challenge

gtopal:
### Description

My deliverable for the GlobalWebIndex Engineering Challenge

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+# Start from the official Golang image
+FROM golang:1.21-alpine AS builder
+WORKDIR /app
+COPY . .
+RUN go build -o server .
+
+FROM alpine:latest
+WORKDIR /app
+COPY --from=builder /app/server .
+EXPOSE 8080
+CMD ["./server"]
--- a/collection.json
+++ b/collection.json
@@ -1,0 +1,145 @@
+    
+    
+{
+  "info": {
+    "name": "GWI Favourites API",
+    "description": "Postman collection for testing GWI favourites endpoints.",
+    "schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
+  },
+  "item": [
+    {
+      "name": "List Favourites",
+      "request": {
+        "method": "GET",
+        "header": [],
+        "url": {
+          "raw": "http://localhost:8080/favourites?user_id=<USER_UUID>",
+          "protocol": "http",
+          "host": ["localhost"],
+          "port": "8080",
+          "path": ["favourites"],
+          "query": [{"key": "user_id", "value": "<USER_UUID>"}]
+        }
+      }
+    },
+    {
+      "name": "Add Chart Favourite",
+      "request": {
+        "method": "POST",
+        "header": [{"key": "Content-Type", "value": "application/json"}],
+        "body": {
+          "mode": "raw",
+          "raw": "{\n  \"type\": \"chart\",\n  \"favorite\": true,\n  \"asset\": {\n    \"ID\": \"<ASSET_UUID>\",\n    \"Title\": \"Sales Q1\",\n    \"XAxisTitle\": \"Month\",\n    \"YAxisTitle\": \"Revenue\",\n    \"Data\": [100, 200, 150],\n    \"Description\": \"Quarter 1 sales chart\"\n  }\n}"
+        },
+        "url": {
+          "raw": "http://localhost:8080/favourites/add?user_id=<USER_UUID>",
+          "protocol": "http",
+          "host": ["localhost"],
+          "port": "8080",
+          "path": ["favourites", "add"],
+          "query": [{"key": "user_id", "value": "<USER_UUID>"}]
+        }
+      }
+    },
+    {
+      "name": "Add Insight Favourite",
+      "request": {
+        "method": "POST",
+        "header": [{"key": "Content-Type", "value": "application/json"}],
+        "body": {
+          "mode": "raw",
+          "raw": "{\n  \"type\": \"insight\",\n  \"favorite\": false,\n  \"asset\": {\n    \"ID\": \"<ASSET_UUID>\",\n    \"Text\": \"40% of millennials spend more than 3 hours on social media daily\",\n    \"Description\": \"Millennial social media usage\"\n  }\n}"
+        },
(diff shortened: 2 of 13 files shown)


Dockerfile
```diff
@@ -0,0 +1,11 @@
+# Start from the official Golang image
+FROM golang:1.21-alpine AS builder
```
alexandros-georgantas, line 2: Could you please explain why you've used this specific version of Go?

collection.json
```diff
@@ -0,0 +1,175 @@
+{
```
alexandros-georgantas, line 1: Nice touch providing postman blueprint

handlers.go
```diff
@@ -0,0 +1,298 @@
+package main
```
alexandros-georgantas, line 1: If you were to provide a bit more structure to this deliverable how you will have done it?

handlers.go
```diff
@@ -0,0 +1,298 @@
...
+	assetIDStr := r.URL.Query().Get("asset_id")
+	assetID, err := uuid.Parse(assetIDStr)
+	if err != nil {
+		http.Error(w, "Invalid asset_id", http.StatusBadRequest)
+		return uuid.UUID{}, false
+	}
+	return assetID, true
+}
+
+func setupRoutes() {
```
alexandros-georgantas, line 36: Cloud you please provide an alternative for the structure of the favorites endpoints (to become more RESTful)?

handlers.go
```diff
@@ -0,0 +1,298 @@
...
+// Create an asset
+func handleAddFavourite(w http.ResponseWriter, r *http.Request) {
+
+	userID := getUserIDFromContext(r)
+	if userID == uuid.Nil {
+		log.Printf("handleAddFavourite: invalid user_id from token")
+		http.Error(w, "Unauthorized", http.StatusUnauthorized)
+		return
+	}
+	user := store.GetUser(userID)
```
alexandros-georgantas, line 102: Lets say that someone asks to change from in-memory storage to a proper database approach. Based on this line, how easy is to make this transition?

models.go
```diff
@@ -0,0 +1,80 @@
...
+import "github.com/google/uuid"
+
+// AssetType represents the type of asset
+const (
+	ChartType    = "chart"
+	InsightType  = "insight"
+	AudienceType = "audience"
+)
+
+type Asset interface {
```
alexandros-georgantas, line 12: Nice

alexandros-georgantas (commented):
Hello @gtopal 👋 ,

Thanks for taking the time working on this assignment!
Your implementation is complete and straight forward covering various aspects.
I have posted some comments in your PR more as a conversation starter. 
To this end, I would appreciate if you find the time to have a look and address my comments!

Thanks,
Alex

gtopal:
> Hello @gtopal 👋 ,
> 
> Thanks for taking the time working on this assignment! Your implementation is complete and straight forward covering various aspects. I have posted some comments in your PR more as a conversation starter. To this end, I would appreciate if you find the time to have a look and address my comments!
> 
> Thanks, Alex

Hello @alexandros-georgantas ,

thanks for the feedback and your time looking my PR. I will reply as soon as possible.

Dockerfile
gtopal: This version is stable, lightweight and reduces the final image size. Moreover, I think it's preferable to use a specific version and not the latest in order not to break anything.

handlers.go
gtopal: For sure some points need refactoring for readability and maintainability purposes.

I will:

- Extract logic into functions:
For example , the creation of the user could go in a seperate function
or
It could be useful also to create a seeder for creating users and assets related to the users and put an appropriate command also in `Makefile`
Another example could be to create a function for starting the server

- Create a `.env.example` and also a `.env` file to capture environmental variables
For example I noticed that the variable `jwtSecret` could be stored in `.env` as `JWT_SECRET`

- Create maybe a seperate folder for tests and put there seperate files for testing, one let's say for API calls and one for more general unit tests

- Put the endpoint calls to a seperate file and the handling logic(functions) to another one.

handlers.go
gtopal: Yes sure.

I know it will be better to structure my endpoints like below:

To add:POST `/users/{user_id}/favourites`
To update: PUT` /users/{user_id}/favourites/{asset_id}`
To delete: DELETE `/users/{user_id}/favourites/{asset_id}`
To list: GET `/users/{user_id}/favourites`

using request parameters like
`http://localhost:8080/favourites/remove?asset_id=c217dd2d-1866-43f0-88a5-5550504fe62d`is not a good practice

handlers.go
gtopal: I am not sure if I understand correct, but I would change only the implementation of the `GetUser` 
The method signature and usage stay the same. The receiver of `GetUser` would change to `*DBStorage` something like `func (s *DBStorage) GetUser(id uuid.UUID) *User { ... }`
and the implementation should query the DB.

In other words, I would take advantage of using **interfaces** in order based on my option to change the implementation based on the receiver.

Of course for the DB declaration I should create a service with an image let's say for Postgresql and define also the connection parameters to `.env`

Dockerfile
alexandros-georgantas: In principal I agree however this version if I am not mistaken is from 2023 https://go.dev/doc/devel/release#go1.21.0

handlers.go
alexandros-georgantas: Here basically I am more interested to hear your thoughts around D.I. and if it is feasible based on how you've implemented your handlers.

Dockerfile
gtopal: ok! Next time I will be aware of not using an obsolete version

handlers.go
gtopal: @alexandros-georgantas As it is , is not feasible cause I have a hard dependency on handlers

I have to define a more flexible approach such as declaring an interface like:

```
type UserStore interface {
    GetUser(id uuid.UUID) *User
}
```
and then make the handlers depend on `UserStore`

```
type Handler struct {
    Store UserStore
}
```

and injecting either an in-memory store or a DB-backed implementation.

PR #42, opened 2025-09-17
Submission of the EM Go coding challenge

A-Ioakeimidis:


README.md
```diff
@@ -1,31 +1,97 @@
...
 
-On the surface the challenge is trivial to solve, however you should choose to add features or capabilities which you feel demonstrate your skills and knowledge the best. For example, you could choose to optimise for performance and concurrency, you could choose to add a robust security layer or ensure your application is highly available. Or all of these.
+## Running locally
+```bash
+go run main.go
+```
 
-Of course, usually we would choose to solve any given requirement with the simplest possible solution, however that is not the spirit of this challenge.
+## API Endpoints
+- `GET /users/{id}` → list favorites
```
alesr, line 17: `GET /users/{id}` would be appropriate for getting users by ID.

README.md
```diff
@@ -1,31 +1,97 @@
...
-* Insight (a small piece of text that provides some insight into a topic, e.g. "40% of millenials spend more than 3hours on social media daily")
-* Audience (which is a series of characteristics, for that exercise lets focus on gender (Male, Female), birth country, age groups, hours spent daily on social media, number of purchases last month)
-e.g. Males from 24-35 that spent more than 3 hours on social media daily.
+## Initial curl commands to test GET, POST, PUT and DELETE apis
 
-Build a web server which has some endpoint to receive a user id and return a list of all the user’s favourites. Also we want endpoints that would add an asset to favourites, remove it, or edit its description. Assets obviously can share some common attributes (like their description) but they also have completely different structure and data. It’s up to you to decide the structure and we are not looking for something overly complex here (especially for the cases of audiences). There is no need to have/deploy/create an actual database although we would like to discuss about storage options and data representations.
+### Create a Chart asset
+```bash
+curl -X POST http://localhost:8080/users/user123 \
```
alesr, line 33: How this API could be designed in a way that I would know that it should create an asset and not a user?

api/apis.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		http.Error(writer, err.Error(), http.StatusBadRequest)
+		return
+	}
+
+	asset.CreatedAt = time.Now()
+
+	err = database.AddFavorite(userId, asset)
+	if err != nil {
+		http.Error(writer, err.Error(), http.StatusInternalServerError)
```
alesr, line 106: What could we do to help the client understand why we failed to favorite the asset?

api/apis_test.go
```diff
@@ -0,0 +1,165 @@
...
+
+	resetData()
+
+	//POST an asset
+	asset := models.Asset{ID: "12", TypeID: "Chart", Description: "Just an asset chart"}
+	body, _ := json.Marshal(asset)
+
+	request, _ := http.NewRequest("POST", "/users/user123/12", bytes.NewBuffer(body))
+	response := httptest.NewRecorder()
+	FavoritesAPIs(response, request)
```
alesr, line 43: Interesting approach using the recorder and passing the request directly to the handler.

Since we’re dealing with an HTTP server though, it might be even cleaner to spin up a test server, register your handlers there, and then make requests against it. That way the test flow looks closer to how the server will actually run in production.

api/apis.go
```diff
@@ -0,0 +1,152 @@
...
+		return
+	}
+
+	writer.Header().Set("Content-Type", "application/json")
+	writer.WriteHeader(http.StatusOK)
+	err = json.NewEncoder(writer).Encode(asset)
+	if err != nil {
+		http.Error(writer, err.Error(), http.StatusInternalServerError)
```
alesr, line 140: Once you use `WriteHeader` in line 137, writing internal error here will be ineffective.

models/asset.go
```diff
@@ -0,0 +1,35 @@
+package models
+
+import (
+	"time"
+)
+
+type Asset struct {
+	ID          string                 `json:"id"`
+	TypeID      string                 `json:"type"`
```
alesr, line 9: Would be possible to create an `Chart` asset with the `TypeID` of the asset set to `Insight`?

alesr (commented):
Thanks a lot for submitting your solution for the assignment! I went through it and left a couple of comments with some interesting findings you might want to take a look at.

A-Ioakeimidis:
> Build a web server which has some endpoint to receive a user id and return a list of all the user’s favourites. Also we want endpoints that would add an asset to favourites, remove it, or edit its description. Assets obviously can share some common attributes (like their description) but they also have completely different structure and data. It’s up to you to decide the structure and we are not looking for something overly complex here (especially for the cases of audiences). There is no need to have/deploy/create an actual database although we would like to discuss about storage options and data representations.
> ### Create a Chart asset
> ```bash
> curl -X POST http://localhost:8080/users/user123 \

 

> Thanks a lot for submitting your solution for the assignment! I went through it and left a couple of comments with some interesting findings you might want to take a look at.

Good afternoon @alesr 
Thank you for taking the time to review the code and leave your feedback. I’ll make sure feedback is addressed without delay.

README.md
A-Ioakeimidis: Right now the design could be confusing because POST /users/{id} usually means creating a user. To avoid and confusion, I would namespace the favorites resource explicitly. For example, POST /users/{id}/favorites which would show a lot more clearly that this is intended to add a favorite asset to the specific user.

api/apis.go
A-Ioakeimidis: In this case, a generic error is thrown in this case a 500. I’d improve this by mapping specific errors to appropriate HTTP status codes. For example, 400 for invalid input, 409 for duplicate assets, and 404 for user not found. I’d also return structured JSON error responses instead of plain text, so the client can programmatically handle the error and show a meaningful message to the user.

api/apis_test.go
A-Ioakeimidis: That’s a good point! I used Recorder because it works well for unit-style tests because it’s lightweight and lets me call the handler directly to quickly test. I would probably use both approaches, handler level tests for speed and coverage, and a few integration style tests using a test server to validate the whole stack.

api/apis.go
A-Ioakeimidis: Thanks for the observation. I’m still getting the hang of Go, so this is really helpful. It’s a good reminder to only write the header after I’m sure encoding succeeded.

models/asset.go
A-Ioakeimidis: Yes, with the current model it’s possible, because TypeID is just a string field and nothing enforces its correctness against the actual struct type. To prevent that, I would need to introduce either validation logic (e.g. check that TypeID == "Chart" when creating a Chart asset), or even redesign the model to avoid duplication, for example, by not storing TypeID separately at all, and instead inferring it from the struct type.

README.md
A-Ioakeimidis: That’s a good point. GET /users/{id} should return the user resource itself. A more RESTful design would be something like GET /users/{id}/favorites to clearly indicate that we’re retrieving the user’s favorite assets. Since this was a small exercise, I simplified the paths, but in a production system I would make that adjustment for clarity and consistency with REST principles.

PR #44, opened 2025-09-26
feat: add endpoints to handle user favorites operations (EM-GO-CHALLENGE)

tmavroeid:
# Summary
Two new endpoints are introduced to support the following operations:
- an endpoint that enables a user to ADD/EDIT(shared attributes)/DELETE favorites
- an endpoint that gets the user id and returns back a list of favorites

## Changes
- Added a `users` table and upgraded `favorites` to store **per-user overrides** (`override` JSONB), so your tweaks don’t change the asset for everyone else.
- Split repos for clarity: `pg.go`, `asset.postgres.go`, `favorites.postgres.go`.
- Favorites still use fast keyset pagination (`after_t` + `after_id`) to keep response time consistent and handle unlimited favorites without burdening the response time.
- Seed to dev locally. Seed creates **2 users**, **9 assets** (3 of each type), and favorites for both.
- Dockerfile and Docker Compose are provided to enable deployment in staging faster (by SRE team) and have the same dev environment between the whole engineering team in order to assist bug resolution in the same environment no matter the OS or other local configuration.


src/infra/http/http.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+// DELETE /v1/users/:user_id/favorites/:asset_id
+func (h *Handlers) removeFavorite(c *gin.Context) {
+	userID := core.UserID(c.Param("user_id"))
+	assetID := core.AssetID(c.Param("asset_id"))
+	if err := h.favoriteSvc.RemoveFromFavorites(userID, assetID); err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": err.Error()})
+		return
+	}
```
pamanta, line 97: Should all errors originating from the service return a bad request? If not, how would you handle different types of errors here and in `addFavorite, patchFavoriteOverride`?

src/infra/http/http.go
```diff
@@ -0,0 +1,145 @@
...
+	c.Status(http.StatusNoContent)
+}
+
+func SetupRouter(h *Handlers) *gin.Engine {
+	r := gin.New()
+	r.Use(gin.Logger(), gin.Recovery())
+	v1 := r.Group("/v1")
+	{
+		v1.GET("/users/:user_id/favorites", h.getFavorites)
+		v1.POST("/users/:user_id/favorites", h.addFavorite)
```
pamanta, line 139: Can you identify any drawbacks in having the `user_id` on the path? If yes how would you handle it in a production environment?

src/scripts/server/main.go
```diff
@@ -0,0 +1,41 @@
...
+	favSvc   := core.NewFavoriteService(assetsRepo, favoritesRepo)
+	now := func() time.Time { return time.Now().UTC() }
+
+	h := apihttp.NewHandlers(assetSvc, favSvc, now)
+	r := apihttp.SetupRouter(h)
+
+	log.Printf("listening on %s", addr)
+	if err := http.ListenAndServe(addr, r); err != nil {
+		log.Fatal(err)
+	}
```
pamanta, line 37: What happens on the in flight requests, when the server receives a signal to shutdown (ex. in kubernetes)? How would you handle this in a production environment

src/scripts/server/main.go
pamanta, line 1: What would be the first thing to test in your application?

src/scripts/server/main.go
```diff
@@ -0,0 +1,41 @@
...
+	// one shared connection pool
+	pg, err := repo.NewPG(dsn)
+	if err != nil { log.Fatal(err) }
+	defer pg.Close()
+
+	// split repos -> interfaces
+	var assetsRepo core.AssetRepository    = repo.NewAssetRepo(pg)
+	var favoritesRepo core.FavoriteRepository = repo.NewFavoriteRepo(pg)
+
+	assetSvc := core.NewAssetService(assetsRepo)
```
pamanta, line 27: this seems to be missing and code is not compiling
```
# platform-go-challenge/src/infra/http
src/infra/http/http.go:14:19: undefined: core.AssetService
src/infra/http/http.go:19:32: undefined: core.AssetService
```

src/scripts/db/migrate.sql
```diff
@@ -0,0 +1,42 @@
+BEGIN;
+
+-- ===== users =====
+CREATE TABLE IF NOT EXISTS users (
+  id          text PRIMARY KEY,
```
pamanta, line 5: could you explain a bit the rationale behind using an id with type `text`?

pamanta (commented):
Hello @tmavroeid,

Thank you very much for taking the time to complete the assessment!

I have also left some comments for you to have a look when you have time :)

PR #45, opened 2025-09-29
Favourite assets challenge

IoannaMitrousi:


server/routes/routes.go
```diff
@@ -0,0 +1,43 @@
...
+	favController *controllers.FavouriteController,
+	authMiddleware func(next http.Handler) http.Handler,
+) {
+	r.Use(authMiddleware)
+
+	// Users
+	r.Route("/users", func(r chi.Router) {
+		r.Post("/", userController.CreateUserHandler)
+		r.Get("/", userController.ListUsersHandler)
+		r.Get("/by-id", userController.GetUserHandler)
```
alexandros-georgantas, line 22: Cloud you please elaborate on this decision of having this `by-id` path and include the id value in a query parameter instead of just having the id as the path of the endpoint e.g. /users/:userId ?

alexandros-georgantas (commented):
Hello @IoannaMitrousi,
Thank you for the taking the time working on this assignment! You did really good!
I have left one comment on your PR so if you find time please come back with an answer.
Also, I would love to hear your rationale behind the decision to use Keycloack.

server/routes/routes.go
IoannaMitrousi: I used /users/by-id?userId=... because I already had GET /users for listing all users, and I wanted to avoid any ambiguity. I know the more RESTful approach would be /users/{userId}, which clearly separates collection vs individual resources. If I were designing it now, I’d go with path parameters for CRUD and keep query parameters for filtering.

IoannaMitrousi:
Hello @alexandros-georgantas 
Thank you so much for your feedback and for taking the time to review my assignment.
I chose to use Keycloak mainly to handle authentication and role-based access control. In the project, users could manage assets and mark favorites, so I needed a way to verify not only that requests were coming from authenticated users but also what roles they had. While the token wasn’t strictly validated in the project, in a real environment Keycloak would perform proper authentication, and the bearer token allows me to confirm the user’s identity. Additionally, by including roles in the token, I could enforce that only admins could perform certain actions, like managing assets. The main goal was to ensure that, for example, when a request includes a userId or assetId, we can verify that the user is actually allowed to access that information.

PR #47, opened 2025-10-07
GWI challenge 

stevenoddity:


middleware/error_handler.go
```diff
@@ -0,0 +1,30 @@
...
+// Parameters:
+//   - next: The next http.Handler in the chain to be executed.
+//
+// Returns:
+//
+//	An http.Handler that wraps the provided handler with recovery functionality.
+func RecoverMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer func() {
+			if err := recover(); err != nil {
```
dzacharakis, line 22: Is recovering from panics in Go generally considered a good practice?

services/user/user_service.go
```diff
@@ -0,0 +1,25 @@
+package user_service
+
+import (
+	"net/http"
+)
+
+// GetUsers handles the HTTP request to retrieve a list of users.
+func GetUsers(w http.ResponseWriter, r *http.Request) {
+	// This function currently does not implement any logic.
```
dzacharakis, line 9: Any reason you’re keeping these?

main.go
```diff
@@ -0,0 +1,15 @@
...
+	"gwi/middleware"
+	"gwi/routes"
+	"log"
+	"net/http"
+)
+
+func main() {
+	router := routes.RegisterRoutes()
+	fmt.Println("🚀 Server running on http://localhost:8080")
+	log.Fatal(http.ListenAndServe(":8080", middleware.RecoverMiddleware(router)))
```
dzacharakis, line 14: Why did you choose `log.Fatal` here? Would `log.Error` have been more appropriate?

utils/input_validators.go
```diff
@@ -0,0 +1,15 @@
...
+	favorite_model "gwi/models/favorite"
+
+	"github.com/go-playground/validator/v10"
+)
+
+var validate = validator.New()
+
+// ValidateFavorite validates the given Favorite struct.
+// It returns an error if the validation fails, otherwise it returns nil.
+func ValidateFavorite(fav *favorite_model.Favorite) error {
```
dzacharakis, line 13: Is this method being used anywhere?

models/asset/asset.go
```diff
@@ -0,0 +1,10 @@
+package asset_model
+
+import "encoding/json"
+
+// Asset represents a digital asset with an ID, associated user ID, and a description.
+type Asset struct {
+	ID          int             `json:"id"`          // Unique identifier for the asset
+	UserID      int             `json:"user_id"`     // Identifier for the user who owns the asset
+	Description json.RawMessage `json:"description"` // Description of the asset
```
dzacharakis, line 9: Using `json.RawMessage` for the Description field is a poor design choice. Why? Any better approaches?

models/favorite/favorite.go
```diff
@@ -0,0 +1,11 @@
+package favorite_model
+
+import asset_model "gwi/models/asset"
```
dzacharakis, line 3: Is an alternative name needed for this import declaration?

services/favorite/favorite_service.go
```diff
@@ -0,0 +1,128 @@
+package favorite_service
+
+import (
+	"encoding/json"
+	asset_model "gwi/models/asset"
+	favorite_model "gwi/models/favorite"
```
dzacharakis, line 6: Don't you find your package structure over-granular? Could you point out a main principle it’s meant to support and how this granularity impacts it?

models/asset/asset.go
```diff
@@ -0,0 +1,10 @@
+package asset_model
```
dzacharakis, line 1: What would you change if this had to grew to, let's say, 20 models?

dzacharakis (commented):
Hi @stevenoddity 👋

Thanks,
Dimitris

main.go
stevenoddity: if it fails, the program cannot recover, so that's why I chose it.

middleware/error_handler.go
stevenoddity: In general no, as it "hides" the issue.

models/asset/asset.go
stevenoddity: I would add a base struct.

models/asset/asset.go
stevenoddity: It hides the structure of description and it makes quering difficult with slow performance. A better approach would be to define a separate structure from the description.

models/favorite/favorite.go
stevenoddity: I found out that `import "gwi/models/asset"` works as well.

services/favorite/favorite_service.go
stevenoddity: My main goal is the separation of concerns: each package should be responsible of one thing and as much decoupled as possible.
This granularity adds too much noise and it's an overkill for such a small app, but I wanted to demonstrate a more generic approach.

services/user/user_service.go
stevenoddity: These were just placeholders for enhancements that I didn't implement.

utils/input_validators.go
stevenoddity: Nope, I didn't utilize it at the end.

stevenoddity (commented):
Hello Dimitris,
Thanks a lot for your time!
I answered your questions and looking forward for a follow-up.

PR #48, opened 2025-10-14
GWI - Go Challenge

GeoRouv:
Hi,

This is my submission for the technical challenge. You can find a detailed guide in the README file.

Let me know if there’s anything you’d like to discuss.

Best,


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# Copy source code
+COPY . .
+
+# Build the binary
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o server ./cmd/server
+
+# Final stage
+FROM alpine:latest
+
```
alexandros-georgantas, line 20: Could you suggest a way of improving the security of the second stage image?

cmd/server/main.go
```diff
@@ -0,0 +1,82 @@
...
+	protectedMux.HandleFunc("GET /api/users/{id}/favorites", handler.GetFavorites(store))
+	protectedMux.HandleFunc("POST /api/users/{id}/favorites", handler.AddFavorite(store))
+	protectedMux.HandleFunc("DELETE /api/users/{id}/favorites/{assetId}", handler.RemoveFavorite(store))
+	protectedMux.HandleFunc("PUT /api/users/{id}/favorites/{assetId}", handler.EditFavorite(store))
+
+	router.Handle("/api/users/", authMiddleware(protectedMux))
+
+	// --- Rate limiting setup ---
+	loginLimiter := middleware.NewRateLimiter(5, time.Minute)     // 5 req/min per IP
+	defaultLimiter := middleware.NewRateLimiter(50, time.Minute) // 10 req/min per IP
```
alexandros-georgantas, line 52: Could you suggest another component of the overall system that could handle the rate limiting functionality?

cmd/server/main.go
```diff
@@ -0,0 +1,82 @@
...
+	rateLimitMiddleware := middleware.RateLimitMiddleware(loginLimiter, defaultLimiter)
```
alexandros-georgantas, line 53: Will your implementation of rate limiting function properly in production where multiple instances of a server will run? Could you suggest any changes?

internal/auth/auth.go
```diff
@@ -0,0 +1,84 @@
...
+	if jm.config.SecretKey == "" {
+		return "", errors.New("secret key not configured")
+	}
+
+	claims := &Claims{
+		UserID: userID,
+		RegisteredClaims: jwt.RegisteredClaims{
+			ExpiresAt: jwt.NewNumericDate(time.Now().Add(jm.config.ExpiryTime)),
+			IssuedAt:  jwt.NewNumericDate(time.Now()),
+			NotBefore: jwt.NewNumericDate(time.Now()),
```
alexandros-georgantas, line 46: Could you please explain the use of this claim? What are you trying to solve with this?

internal/handler/auth.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		// Generate token
+		token, err := jwtManager.GenerateToken(req.UserID)
+		if err != nil {
+			util.RespondError(w, http.StatusInternalServerError, "Failed to generate token")
+			return
+		}
+
+		util.RespondJSON(w, http.StatusOK, map[string]interface{}{
```
alexandros-georgantas, line 35: Could you please explain what is the value of this util? Also, if you had time to refactor this part and provide a bit more structure what you would have done differently in terms of the response?

cmd/server/main.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// Public endpoints (no auth required)
+	router.HandleFunc("GET /api/health", handler.HealthCheck(metricsCollector))
+	router.HandleFunc("POST /api/login", handler.Login(jwtManager))
+
+	// Protected endpoints (auth required)
+	authMiddleware := middleware.AuthMiddleware(jwtManager)
+
+	protectedMux := http.NewServeMux()
+	protectedMux.HandleFunc("GET /api/users/{id}/favorites", handler.GetFavorites(store))
```
alexandros-georgantas, line 43: Could you please explain your decision on passing the data layer directly to the API handlers? Could you think of any limitations to this approach? If you had more time how would you refactor it?

internal/model/asset.go
```diff
@@ -0,0 +1,80 @@
...
+
+// Supported asset types
+const (
+	AssetTypeChart    AssetType = "chart"
+	AssetTypeInsight  AssetType = "insight"
+	AssetTypeAudience AssetType = "audience"
+)
+
+// BaseAsset contains common fields for all assets
+type BaseAsset struct {
```
alexandros-georgantas, line 14: Could you think of a different type that would perhaps be more suitable for this ?

internal/storage/store.go
```diff
@@ -0,0 +1,116 @@
...
+
+func NewMemoryStore() *MemoryStore {
+	return &MemoryStore{
+		users: make(map[string]*model.User),
+	}
+}
+
+// Returns all favorites for a user
+func (ms *MemoryStore) GetUserFavorites(userID string) ([]*model.Favorite, error) {
+	ms.mu.RLock()
```
alexandros-georgantas, line 33: Could you explain your rationale behind this lock for this read operation?

internal/handler/auth.go
```diff
@@ -0,0 +1,40 @@
...
+		}
+
+		if req.UserID == "" {
+			util.RespondError(w, http.StatusBadRequest, "user_id is required")
+			return
+		}
+
+		// Generate token
+		token, err := jwtManager.GenerateToken(req.UserID)
+		if err != nil {
```
alexandros-georgantas, line 30: Could you please explain what will happen to any error msgs thrown by the `GenerateToken`?

internal/storage/store.go
```diff
@@ -0,0 +1,116 @@
...
+	favorites := make([]*model.Favorite, 0, len(user.Favorites))
+	for _, fav := range user.Favorites {
+		favorites = append(favorites, fav)
+	}
+
+	return favorites, nil
+}
+
+// Adds a new asset to user's favorites
+func (ms *MemoryStore) AddFavorite(userID string, asset *model.Asset) error {
```
alexandros-georgantas, line 50: Could you please explain what will happen if I try to add the same favorite more than once?

alexandros-georgantas (commented):
Hello @GeoRouv !

Thank you for your time and effort implementing this assignment! Your deliverable was in a good shape however, I faced some issues when it comes to the JWT secret.

I've left some comments on your PR and I would really appreciate it if you could spend some time answering them.

Thanks,
Alex

internal/model/asset.go
```diff
@@ -0,0 +1,80 @@
...
+	ID          string    `json:"id"`
+	Type        AssetType `json:"type"`
+	Description string    `json:"description"`
+	CreatedAt   int64     `json:"created_at"`
+	UpdatedAt   int64     `json:"updated_at"`
+}
+
+// Chart represents a chart asset
+type Chart struct {
+	BaseAsset `json:",inline"`
```
dzacharakis, line 24: What’s the purpose of this option in the field tag?

internal/model/asset.go
```diff
@@ -0,0 +1,80 @@
...
+	Countries []string `json:"countries"`
+	AgeMin    int      `json:"age_min"`
+	AgeMax    int      `json:"age_max"`
+	MinHours  int      `json:"min_hours_daily_social_media"`
+	MinBuys   int      `json:"min_purchases_last_month"`
+}
+
+// Asset is a union type that holds any asset
+type Asset struct {
+	Chart    *Chart    `json:"chart,omitempty"`
```
dzacharakis, line 51: Suppose the following valid asset definition:
(I also took a look at your similar test case in `asset_test.go` line 79).

```go
asset := &model.Asset{
    Chart: &model.Chart{
        BaseAsset: model.BaseAsset{
            ID: "chart-1",
            Type: model.AssetTypeChart,
        },
    },
    Audience: &model.Audience{
        BaseAsset: model.BaseAsset{
            ID: "audience-1",
            Type: model.AssetTypeAudience,
        },
    },
    Insight: &model.Insight{
        BaseAsset: model.BaseAsset{
            ID: "insight-1",
            Type: model.AssetTypeInsight,
        },
    },	
}
```

The POST `/api/users/{id}/favorites` controller receives a request with this in its request body. This asset will be stored as favorited by user {id}.

How many favorite assets does this user have? What are your thoughts on this approach? Does it achieve the desired behavior or address the underlying issue?

internal/handler/handler.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// RemoveFavorite removes an asset from user's favorites
+func RemoveFavorite(store storage.Store) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		userID := r.PathValue("id")
+		assetID := r.PathValue("assetId")
+
+		if err := store.RemoveFavorite(userID, assetID); err != nil {
+			if err == api_errors.ErrUserNotFound || err == api_errors.ErrFavoriteNotFound {
```
dzacharakis, line 84: Here `ErrUserNotFound`, `ErrFavoriteNotFound` are imported from the `platform-go-challenge/internal/api_errors` package. 

Suppose two new directories, `app` and `mainpkg`, are added to the repository so that:
- the same package resides at `platform-go-challenge/app/internal/api_errors`, and 
- a new package is introduced at `platform-go-challenge/mainpkg`

In the `source.go` file under the `mainpkg` package, you still need to use these two variables.
If feasible, specify the appropriate import path required for this configuration to work correctly.

dzacharakis (commented):
Hi @GeoRouv 👋

Thanks,
Dimitris

cmd/server/main.go
GeoRouv: It could be handled by an external component such as an API gateway or reverse proxy

cmd/server/main.go
GeoRouv: Right now, if there are multiple server instances, each instance tracks requests independently, so clients could bypass the limit by hitting different servers. We could use one of these (or both) instead: 
- An external cache (Redis) to share all rate-limit state across all instances
- An API Gateway / Reverse Proxy (Nginx, Cloudflare, AWS API Gateway e.t.c) to enforce rate limits across all instances

internal/auth/auth.go
GeoRouv: This claim makes the JWT:
- Carry user identity and validity info
- Expire automatically without needing a server-side session
- Stateless / no need to store active sessions in a database

In short, it defines _who_ the token belongs to and _when_ it’s valid.

internal/handler/auth.go
GeoRouv: Some of the key points are:

- Centralized response formatting
- All endpoints return same JSON structure
- Single place to handle encoding errors


About the second part of the question, I can see that `map[string]interface{}` might be a bit vague by lacking type safety and consistency. Instead, I would replace it with typed structs, a standardize envelope e.g. {success, data, error} and error codes for better client handling.

cmd/server/main.go
GeoRouv: The direct storage-to-handler makes testing difficult and mixes concerns therefore I'd add a service layer between handlers and storage for better separation of concerns, easier mocking in tests, and flexibility to add business logic, caching (Redis as I talked about earlier), or transactions without touching handlers

internal/model/asset.go
GeoRouv: I'd switch to an interface approach where `Asset` is just an interface with methods like `GetID() `and `GetType()`, and Chart, Insight, and Audience each implement those methods. This will get rid of the union type and switch statements, and will make it easier to add new asset types

internal/storage/store.go
GeoRouv: The read lock prevents data races when concurrent requests access. Without it, reading while another operation is writing could cause panics or corrupted data. Essentially, it allows multiple concurrent reads for better performance while still blocking writes

internal/handler/auth.go
GeoRouv: The original error from `GenerateToken` is discarded. It's not logged or returned anywhere. This will make debugging issues impossible since you only see the generic "Failed to generate token" message. I'd add logging here to capture the actual error details while still returning the generic message to the client

internal/storage/store.go
GeoRouv: Adding the same favorite twice will silently overwrite the existing one without returning an error. This means the original `AddedAt` timestamp gets lost, and the user has no way to know it was a duplicate. I'd add a check before inserting to return an error if the favorite already exists

internal/model/asset.go
GeoRouv: It merges the embedded struct's fields directly into the parent JSON object instead of creating a nested object. This gives a flatter, cleaner API response structure

internal/model/asset.go
GeoRouv: The user would have 1 favorite containing all 3 asset types at the same time, which is invalid. The `Asset` union type has no validation to enforce that exactly one type should be set. `GetBaseAsset()` would just return the first non-nil field (`Chart` in this case), silently ignoring the others. I'd use the interface-based approach which makes invalid states impossible by design

internal/handler/handler.go
GeoRouv: `mainpkg` at the top level cannot import `platform-go-challenge/app/internal/api_errors ` because internal packages can only be imported by code within the same parent tree.

We could use one of these:

1. Move `mainpkg` inside `app/ ` so it becomes `platform-go-challenge/app/mainpkg/`, then use:

` import "platform-go-challenge/app/internal/api_errors"`

2. Move `api_errors` outside of `internal/` to `platform-go-challenge/app/api_errors/`, then use:

`import "platform-go-challenge/app/api_errors"`

GeoRouv:
Hello @alexandros-georgantas & @dzacharakis ! 🥷 Thank you for taking the time to review my implementation and share your feedback. I’ve added some comments below your points. Looking forward to hearing your thoughts.

Dockerfile
GeoRouv: A minimal base image with no shell or package manager, accompanied with only the necessary packages can reduce the attack surface

alexandros-georgantas: i was looking for something that has to do with the users of the containers

internal/auth/auth.go
alexandros-georgantas: thanks for the answer but I am not referring to the claims usability in general. I am more interested to know why you've chosen the `NotBefore` claim.

Dockerfile
GeoRouv: Ah I see now. I'd create a dedicated non-root user to run the container in order to have the least privileges

internal/auth/auth.go
GeoRouv: It ensures the token is not considered valid until a specific time. It can be seen as a grace period before token usage in case additional checks or actions need to be done. Potentially, it can also prevent some attacks/time-manipulations of the token

internal/model/asset.go
dzacharakis: From what I see (anonymous field), `BaseAsset` is embedded within `Chart`, so I’m wondering if the tag is necessary.

GeoRouv: Indeed, it's redundant

PR #50, opened 2025-10-31
GWI-Challenge

MichailidouNatalia:


preferred_assets_api/internal/adapters/repositories/inmemory/favourite_lru_repository.go
```diff
@@ -0,0 +1,141 @@
...
+	"github.com/MichailidouNatalia/GWI-Engineering-Challenge/preferred_assets_api/internal/adapters/repositories/entities"
+	"github.com/MichailidouNatalia/GWI-Engineering-Challenge/preferred_assets_api/internal/ports"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+var _ ports.FavouriteRepository = (*LRUFavouriteRepositoryImpl)(nil)
+
+type LRUFavouriteRepositoryImpl struct {
+	userAssetsCache *lru.Cache[string, map[string]time.Time]
+	existsCache     *lru.Cache[string, bool]
```
lkslts64, line 16: I notice you have two caches here: userAssetsCache and existsCache. Can you walk me through your reasoning for having both? What problem does the existsCache solve?

preferred_assets_api/internal/adapters/repositories/inmemory/favourite_lru_repository.go
```diff
@@ -0,0 +1,141 @@
...
+
+	mu sync.RWMutex
```
lkslts64, line 18: Nice use of RWMutex 💯

preferred_assets_api/internal/adapters/http/handlers/user_handler.go
```diff
@@ -0,0 +1,260 @@
...
+
+	u, err := h.service.GetUserByID(id)
+	if err != nil {
+		http.Error(w, "user not found", http.StatusNotFound)
+		return
+	}
+
+	usr := mapping.DomainToUserRes(*u)
+	if err := json.NewEncoder(w).Encode(usr); err != nil {
+		log.Fatal(err)
```
lkslts64, line 98: Can you spot any issues in the way this error is handled?

preferred_assets_api/internal/adapters/repositories/inmemory/asset_lru_repository.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (r *LRUAssetRepositoryImpl) Save(asset entities.AssetEntity) (entities.AssetEntity, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	if err := asset.Validate(); err != nil {
+		return entities.AssetBaseEntity{}, err
+	}
+
+	val, ok := r.cache.Get(asset.GetID())
```
lkslts64, line 44: I would expect the save method to create an asset, not just retrieve it. Is that right?

lkslts64 (commented):
Hey @MichailidouNatalia, thanks for submitting the challenge! We really appreciate your effort. 

You did a great job, following a clean architecture and designing a solid RESTful API. Furthermore, you added a lot of "extras", such as Open API docs, RBAC, JWT Auth, Docker and lots of unit tests 💯 

Could you please take a look at the comments I've left below?

preferred_assets_api/internal/adapters/repositories/inmemory/favourite_lru_repository.go
MichailidouNatalia: It is used for quick search validation during favourites addition through the API endpoint and it could also be used for other optimized searches where you don't need to fetch the entire object. It is a bit overengineering I have to admit for that small objects like this, but it would be very useful for fast searches if we are talking about very large datasets as described in the assignment.

preferred_assets_api/internal/adapters/http/handlers/user_handler.go
MichailidouNatalia: Yes, totally my bad. It should be handled in a graceful manner returning an Internal Server Error status and an appropriate message. Using Fatal is far too sever and would close the entire application.

preferred_assets_api/internal/adapters/repositories/inmemory/asset_lru_repository.go
MichailidouNatalia: You are right, the copy-paste demon from another method. 
It should be like this 
```
func (r *LRUAssetRepositoryImpl) Save(asset entities.AssetEntity) (entities.AssetEntity, error) {
	r.mu.Lock()
	defer r.mu.Unlock()

	if err := asset.Validate(); err != nil {
		return entities.AssetBaseEntity{}, err
	}
	r.cache.Add(asset.GetID(), asset)

	return asset, nil
}
```

README.md
```diff
@@ -1,31 +1,237 @@
...
+### 3. Insight
+Text-based insights with:
+- Descriptive text
+- Key findings and observations
+
+## Quick Start
+
+### Prerequisites
+- Docker and Docker Compose
+- Go 1.19+ (if running locally)
```
alesr, line 81:
```bash
GWI-Engineering-Challenge/p/c/a/main main ❱ go run main.go
go: ../../../go.mod requires go >= 1.25.3 (running go 1.24.2; GOTOOLCHAIN=local)
```

https://github.com/GlobalWebIndex/platform-go-challenge/pull/50/files#diff-a1db9be666bcdbe5b538ebc6b846b0884a19343604fefcaecb7f0bb678b76604R3

preferred_assets_api/internal/adapters/http/handlers/asset_handler.go
```diff
@@ -0,0 +1,110 @@
...
+// @Param request body dto.AssetRequest true "Asset creation request"
+// @Success 201 {object} dto.AssetCreationResponse
+// @Failure 400 {string} string "Invalid input data"
+// @Failure 405 {string} string "Method not allowed"
+// @Failure 500 {string} string "Internal server error"
+// @Security BearerAuth
+// @Router /assets [post]
+func (h *AssetHandler) Create(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, "method not allowed", http.StatusMethodNotAllowed)
```
alesr, line 40: [the post here](https://github.com/MichailidouNatalia/GWI-Engineering-Challenge/blob/main/preferred_assets_api/cmd/api/server/server.go#L102) should be enough, no?

preferred_assets_api/internal/adapters/http/middleware/requests_validator.go
```diff
@@ -0,0 +1,78 @@
...
+			}
+
+			// Validate struct fields using tags
+			if err := validate.Struct(body); err != nil {
+				http.Error(w, err.Error(), http.StatusBadRequest)
+				return
+			}
+
+			// Add the validated body to context
+			ctx := context.WithValue(r.Context(), validatedBodyKey, body)
```
alesr, line 59: I really like the idea of introducing middleware for request validation, and the implementation looks clean as well! That said, I have a few questions:

1. `ValidateBody[T any]`: I’m curious about your decision to use a generic implementation while still allowing `T` to be unconstrained. 

Have you considered how applying a [type constraint](https://go.dev/blog/intro-generics#type-sets) might have influenced your design?

2. Why do we need to pass the validated body as a value in the context? From a validation middleware, I’d typically expect either an error or a boolean flag indicating whether the data is valid.

3. Are there any downsides to using context to pass business data around?

preferred_assets_api/internal/adapters/repositories/entities/asset_entity.go
```diff
@@ -0,0 +1,46 @@
...
+
+type AssetEntity interface {
+	GetID() string
+	GetType() AssetType
+	GetTitle() string
+	GetDescription() string
+	GetCreatedAt() time.Time
+	GetUpdatedAt() time.Time
+	Validate() error
+}
```
alesr, line 25: I would have expected the different asset types (audience, chart, and insights) to implement the Asset interface directly, rather than doing so implicitly through the anonymous embedding of AssetBaseEntity.

As it stands, we could create any arbitrary struct, embed `AssetBaseEntity`, and it would be treated as a valid asset by the type system. In that case, we would still need to rely on runtime type assertions in the mapper rather than leveraging Go's compile-time type safety.

When I think about objects implementing an interface, I generally expect that the goal is to enable different types to share common behavior, where the key differences lie in their specific validation logic and business rules, not in a shared embedded base struct.

Have you considered defining an interfaces that each asset type could explicitly implement instead of relying on anonymous embedding to satisfy the interface indirectly? If so, what trade-offs did you identify between these approaches? Or, did you considered any alternative approach to the assets handling?

alesr (commented):
Hi @MichailidouNatalia 👋🏼 Thank you so much for your assignment submission! I really enjoyed your restful API and the way you implemented the hexagonal architecture, it’s lovely work.

I’d love to hear your thoughts on the points I raised. Also, could you please take a moment to review the instructions for running your application? We encountered some issues running it both locally and via Docker.

Image

Thanks again!

preferred_assets_api/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+      context: .
+      dockerfile: Dockerfile-keycloak
+    environment:
+      - KEYCLOAK_ADMIN=admin
+      - KEYCLOAK_ADMIN_PASSWORD=admin
+      - KC_HEALTH_ENABLED=true
+    ports:
+      - "8090:8080"
+    volumes:
+      - ~./untitled folder/GWI-Engineering-Challenge/preferred_assets_api/cmd/api/config/keycloak-realm-import.json:/opt/keycloak/data/import/realm.json
```
MichailidouNatalia, line 38: @alesr can you please replace the volumes file with this:
`~./preferred_assets_api/cmd/api/config/keycloak-realm-import.json:/opt/keycloak/data/import/realm.json`

preferred_assets_api/docker-compose.yml
```diff
@@ -0,0 +1,49 @@
...
+    tmpfs:
+      - /tmp:size=512M
+    command: start-dev --features=scripts --import-realm --http-enabled=true
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8443/health/ready"]
```
MichailidouNatalia, line 43: @alesr please replace this `http://localhost:8443/health/ready` with `http://localhost:8080/health/ready`

preferred_assets_api/internal/adapters/http/handlers/asset_handler.go
MichailidouNatalia: I admit it's over-defensive programming. With the proper CORS policy set up it can be omitted.

README.md
MichailidouNatalia: It was a last minute addition using AI and did not check if it used the correct versions.
 1.25.3 is the correct version and my Docker setup uses the latest (1.25.4)

preferred_assets_api/internal/adapters/http/middleware/requests_validator.go
MichailidouNatalia: I spent sometime reading about this and I see your point. 
It seems that Go prefers a more explicit approach in comparison to .NET (which is my basic background).

1. To me is a more flexible way to use requests_validator middleware.
We could add an interface to make it more compile-time safe.

2. I tried to avoid parsing the body twice, another solution would be to use a a handler wrapper passing back the validated parsed body. 

3. Using context like this seems to be a Go anti-pattern and I get the reasoning behind this.
 - It's not type-safe
 - It's more complex to read and test
 - It is used mainly for request-scoped values (cancellation, timeouts, deadlines) and not business data.

preferred_assets_api/internal/adapters/repositories/entities/asset_entity.go
MichailidouNatalia: Another implementation that is .NET inspired. 😅 

My approach here was based on the class-base inheritance which as I understand now is not preferred in Go.

I was mainly aiming to reduce boilerplate code. I can see that the most idiomatic approach would be to to explicitly implement the interface for each type, which would benefit of a more type safe coding and simpler clearer architecture.

MichailidouNatalia (commented):
Thank you both @alesr and @lkslts64 for the excellent questions! This was great food for thought and provided a valuable opportunity to revisit my code while diving deeper into Go's design principles.
Let me know if the Docker issues were resolved following my suggestions, and please don't hesitate to reach out if there are any other topics you'd like to discuss.

PR #53, opened 2025-11-25
Implementation: GlobalWebIndex Engineering Challenge

gioannid:
Dear GWI team,

please find hereby my implementaion of GWI Engineering Challenge for your review. Looking forward for your comments!

With regards,
George


alesr:
Hi @gioannid! Thanks for sharing your assignment solution with us. We really appreciate how easy it is to test. We’re currently reviewing your code and will get back to you as soon as possible 🙇🏼

internal/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+package config
+
+import (
+	"os"
+	"strconv"
+	"time"
+)
+
+// Singleton instance of configuration
+var appConfig *Config = nil
```
dzacharakis, line 11:
```suggestion
var appConfig *Config
```

Is there any difference?

internal/config/config.go
```diff
@@ -0,0 +1,101 @@
...
+		IdleTimeout:   getDurationEnv("IDLE_TIMEOUT", 60*time.Second),
+		MaxPageItems:  getIntEnv("MAX_PAGE_ITEMS", 100),
+		AuthEnabled:   getBoolEnv("AUTH_ENABLED", false),
+		// TODO dummy JWT_SECRET value for development; in production use a secure, random secret of at least 256 bits
+		// Below secret along with following data:
+		// {"user_id":"a0eebc99-9c0b-4ef8-bb6d-6bb9bd380a11","exp":1855920000}
+		// corresponds to HS256 JWT token (valid until 23rd Oct 2028):
+		// eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ1c2VyX2lkIjoiYTBlZWJjOTktOWMwYi00ZWY4LWJiNmQtNmJiOWJkMzgwYTExIiwiZXhwIjoxODU1OTIwMDAwfQ.2OG1_i188KTGnOYAN7RCNpCEzY-nmwdO5UdLTfHSWOU
+		// If AUTH_ENABLED, use this token in all /api/v1 requests.
+		JWTSecret: getEnv("JWT_SECRET", "a-string-secret-at-least-256-bits-long"),
```
dzacharakis, line 52: How would you manage this secret in a production-ready application?

internal/domain/asset.go
```diff
@@ -0,0 +1,125 @@
...
+	AssetTypeInsight  AssetType = "insight"
+	AssetTypeAudience AssetType = "audience"
+)
+
+// Asset represents a generic asset that can be favourited
+type Asset struct {
+	ID          uuid.UUID       `json:"id"`
+	Type        AssetType       `json:"type"`
+	Description string          `json:"description"`
+	Data        json.RawMessage `json:"data" swaggertype:"object,string" example:"{\"title\":\"Sample Chart\"}"` // Polymorphic data field
```
dzacharakis, line 26: You’re using `json.RawMessage` for this polymorphic field. What are the drawbacks of this approach? Can you suggest a better alternative?

internal/server/server.go
```diff
@@ -0,0 +1,104 @@
...
+
+	// Conditionally apply JWT authentication middleware *only* to the API v1 subrouter
+	if cfg.AuthEnabled {
+		api.Use(middleware.JWTAuth(cfg.JWTSecret))
+	}
+
+	// Asset management (these handlers will be protected if auth is enabled)
+	api.HandleFunc("/assets", h.CreateAsset).Methods(http.MethodPost)
+	api.HandleFunc("/assets", h.ListAssets).Methods(http.MethodGet)
+	api.HandleFunc("/assets/{assetId}/description", h.UpdateAssetDescription).Methods(http.MethodPatch)
```
dzacharakis, line 68: For the endpoint `/assets/{assetId}/description`, what was your reasoning for using `PATCH` rather than another HTTP method?

internal/repository/memory/memory_test.go
```diff
@@ -0,0 +1,630 @@
...
+		time.Sleep(10 * time.Millisecond)
+
+		asset3, _ := domain.NewAsset(domain.AssetTypeAudience, "Third Audience", domain.AudienceData{
+			Gender: "Male", BirthCountry: "USA", AgeGroups: []string{"18-24"},
+			HoursSocialDaily: 3.0, PurchasesLastMonth: 5,
+		})
+
+		repo.CreateAsset(ctx, asset1)
+		repo.CreateAsset(ctx, asset2)
+		repo.CreateAsset(ctx, asset3)
```
dzacharakis, line 360: `repo.CreateAsset` might return an error but it doesn’t seem to be checked here

internal/repository/memory/memory_test.go
```diff
@@ -0,0 +1,630 @@
...
+		time.Sleep(10 * time.Millisecond)
+
+		asset2, _ := domain.NewAsset(domain.AssetTypeChart, "Chart 2", domain.ChartData{
+			Title: "Chart", AxisXTitle: "X", AxisYTitle: "Y", Data: [][]float64{{1, 2}},
+		})
+		repo.CreateAsset(ctx, asset2)
+		time.Sleep(10 * time.Millisecond)
+
+		// Update asset1 (making it the most recently updated)
+		repo.UpdateAssetDescription(ctx, asset1.ID, "Updated Chart 1")
```
dzacharakis, line 529: also here `repo.UpdateAssetDescription` error return value is not checked

test/api_test.go
```diff
@@ -0,0 +1,451 @@
...
+		time.Sleep(10 * time.Millisecond) // Ensure different timestamps
+	}
+
+	// Test pagination: page 1
+	resp, err := http.Get(ts.URL + "/api/v1/users/" + userID.String() + "/favourites?limit=5&offset=0")
+	require.NoError(t, err)
+	assert.Equal(t, http.StatusOK, resp.StatusCode)
+
+	var page1 handler.Response
+	json.NewDecoder(resp.Body).Decode(&page1)
```
dzacharakis, line 254: and here `(*encoding/json.Decoder).Decode`

internal/handler/handler.go
```diff
@@ -0,0 +1,125 @@
...
+type HealthResponse struct {
+	Success bool                   `json:"success" example:"true"`
+	Data    map[string]interface{} `json:"data" swaggertype:"object,string" example:"{\"status\":\"ok\"}"`
+}
+
+// respondJSON sends a JSON response
+func respondJSON(w http.ResponseWriter, statusCode int, payload interface{}) {
+	w.Header().Set("Content-Type", "application/json")
+	w.WriteHeader(statusCode)
+	json.NewEncoder(w).Encode(payload)
```
dzacharakis, line 85: also here `(*encoding/json.Encoder).Encode`. 
Is there a way to find easily errors like this and improve code quality?

internal/handler/favourite.go
```diff
@@ -0,0 +1,384 @@
+package handler
+
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+
+	"github.com/gioannid/platform-go-challenge/internal/domain"
```
dzacharakis, line 8: Given:

```
platform-go-challenge/
  app/
    internal/
      domain/
        asset.go              // package domain, exports AssetTypeChart
    mainpkg/
      source.go               // package mainpkg
```

In `source.go`, we want to use `domain.AssetTypeChart`.

Is it legal to import `app/internal/domain` from `mainpkg`? Why? If yes, what import path would you use?
If we move `mainpkg` up to `platform-go-challenge/mainpkg`, would your answer change? Why?

internal/domain/asset.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"time"
+
+	"github.com/google/uuid"
+)
+
+// AssetType represents the type of asset
+type AssetType string
```
dzacharakis, line 13: Can you think of a type that might suit this more general use case even better?

internal/handler/favourite.go
```diff
@@ -0,0 +1,384 @@
...
+//	 @Security BearerAuth
+//		@Param			userId	path		string				true	"User ID (UUID)"
+//		@Param			request	body		AddFavouriteRequest	true	"Favourite details"
+//		@Success		201		{object}	Response{data=domain.Favourite}
+//		@Failure		400		{object}	BadRequestError
+//		@Failure		404		{object}	NotFoundError
+//		@Failure		409		{object}	ConflictError
+//		@Failure		500		{object}	InternalServerError
+//		@Router			/users/{userId}/favourites [post]
+func (h *Handler) AddFavourite(w http.ResponseWriter, r *http.Request) {
```
dzacharakis: Could you please explain what will happen if I try to add the same favorite more than once?

dzacharakis (commented):
Hi @gioannid 👋

Thanks,
Dimitris

internal/config/config.go
gioannid: Not really, pointer will anyhow default to nil upon declaration... why it got there? I'm guessing muscle memory from Python where one has to actually assign value to a variable in order to declare it and access it for the first time before lazy-initialize it.

internal/config/config.go
gioannid: The easiest acceptable way would be to just remove the default secret from the initialization (line 52) and depend entirely on its availability as an environment variable JWT_SECRET. If using Docker Compose, the secret could remain entirely off the codebase and be automatically injected in the container by manually or automatically (e.g. by CI/CD machinery) editing the compose.yaml's enviroment block accordingly. (Similar to AUTH_ENABLED already there.) For even more complex production environments like Kubernetes there are dedicated secret stores (e.g. Vault, I think also K8s has its own native secrets object) that could undertake the job of injecting the secret without having it present even in the deployment charts.

internal/domain/asset.go
gioannid: > // RawMessage is a raw encoded JSON value.
> // It implements [Marshaler] and [Unmarshaler] and can
> // be used to delay JSON decoding or precompute a JSON encoding.
> type RawMessage []byte

So this type is just an opaque []byte array that during runtime is accepting the marshalled json structure provided to NewAsset(), then checked against proper schema in another dedicated routine Validate(). The biggest drawback therefore is that we are completely skipping the Go's type checking and its accompanying benefits (static analysis during coding, compile vs. run time failure, early detection of schema incompatibilities closer to the bug source etc.) Also we have introduced boilerplate code to manually check against the three different Data schemas possible. 

An alternative handling would be to introduce to its place a new interface AssetData defining the Validate() method and three overloaded Validate() methods. (Basically along with the 3 existing *Data structures, to construct ourselves the proper class hierarchy that would come much more naturally in a language like Java...) NewAsset() then would accept an AssetData "descedant" and we would have to implement only the business logic validation behind each field.

internal/server/server.go
gioannid: The rationale is that the "/assets/{assetId}/description" endpoint, could be considered as a part of the underlying object (the description field belonging to the asset resource). With the PATCH method we are emphasizing that here we are applying a part edit of the resource vs. a complete substitution which could be e.g. a PUT method on the /assets/{assetId}. The difference may not be important in a toy project like this and of course PUT on the description would work fine as well, but the partial nature of the edit could be good to be reflected to the API for more complex schemas and data access cases. E.g. it could signal that there are no atomicity or consistency guarantees in case of simultaneous edits to the same resource.

internal/repository/memory/memory_test.go
gioannid: Correct. We need to check at least for this case of MemoryRepository.CreateAsset():
```
	if _, exists := r.assets[asset.ID]; exists {
		return domain.ErrAlreadyExists
	}

```

internal/repository/memory/memory_test.go
gioannid: Arguably not needed (the "argue" here refers to the discussion whether it is better to fill with checks the testcase code in order to detect more targeted each failure..). If we force the UpdateAssetDescription() to go through the `return domain.ErrNotFound` path, the testcase fails:
```
=== RUN   TestMemoryRepository_ListAssets/sorts_by_updated_at
    r:\cv\repo\platform-go-challenge\internal\repository\memory\memory_test.go:537:
                Error Trace:    r:/cv/repo/platform-go-challenge/internal/repository/memory/memory_test.go:537
                Error:          Not equal:
                                expected: uuid.UUID{0x92, 0xbd, 0xcd, 0xfc, 0x0, 0xa6, 0x4e, 0x8, 0xbd, 0xe1, 0xf, 0x65, 0x88, 0x38, 0xb6, 0xa7}
                                actual  : uuid.UUID{0x91, 0xcd, 0x5d, 0x9, 0xb5, 0xf9, 0x44, 0x3c, 0x85, 0x6c, 0xfe, 0xb7, 0x37, 0x35, 0x10, 0x33}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 (uuid.UUID) (len=16) {
                                - 00000000  92 bd cd fc 00 a6 4e 08  bd e1 0f 65 88 38 b6 a7  |......N....e.8..|
                                + 00000000  91 cd 5d 09 b5 f9 44 3c  85 6c fe b7 37 35 10 33  |..]...D<.l..75.3|
                                 }
                Test:           TestMemoryRepository_ListAssets/sorts_by_updated_at
    r:\cv\repo\platform-go-challenge\internal\repository\memory\memory_test.go:538:
                Error Trace:    r:/cv/repo/platform-go-challenge/internal/repository/memory/memory_test.go:538
                Error:          Not equal:
                                expected: uuid.UUID{0x91, 0xcd, 0x5d, 0x9, 0xb5, 0xf9, 0x44, 0x3c, 0x85, 0x6c, 0xfe, 0xb7, 0x37, 0x35, 0x10, 0x33}
                                actual  : uuid.UUID{0x92, 0xbd, 0xcd, 0xfc, 0x0, 0xa6, 0x4e, 0x8, 0xbd, 0xe1, 0xf, 0x65, 0x88, 0x38, 0xb6, 0xa7}

                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,3 @@
                                 (uuid.UUID) (len=16) {
                                - 00000000  91 cd 5d 09 b5 f9 44 3c  85 6c fe b7 37 35 10 33  |..]...D<.l..75.3|
                                + 00000000  92 bd cd fc 00 a6 4e 08  bd e1 0f 65 88 38 b6 a7  |......N....e.8..|
                                 }
                Test:           TestMemoryRepository_ListAssets/sorts_by_updated_at
--- FAIL: TestMemoryRepository_ListAssets/sorts_by_updated_at (0.03s)
=== RUN   TestMemoryRepository_ListAssets/concurrent_read_access_is_thread-safe
--- PASS: TestMemoryRepository_ListAssets/concurrent_read_access_is_thread-safe (0.00s)
=== RUN   TestMemoryRepository_ListAssets/partial_page_at_end_of_results
--- PASS: TestMemoryRepository_ListAssets/partial_page_at_end_of_results (0.00s)
--- FAIL: TestMemoryRepository_ListAssets (0.08s)
FAIL
FAIL    github.com/gioannid/platform-go-challenge/internal/repository/memory    0.452s
```

test/api_test.go
gioannid: I'm counting on the panic that will follow right in the next line `data1 := page1.Data.(map[string]interface{})` :)

internal/handler/handler.go
gioannid: A static analysis tool as part of my IDE could probably catch and highlight cases where there is no error check done at least for known library functions.

internal/handler/favourite.go
gioannid: In general, we can import "internal" resources only within the tree structure of the "internal" parent (here: app). So in the above structure we could import `"platform-go-challenge/app/internal/domain"` from mainpkg under app.
If we move mainpkg just under platform-go-challenge, then it is a sibling to the app directory so Go compiler would reject the import.

internal/domain/asset.go
gioannid: Right. We could try to use the closest that Go has to enums:

```
type AssetType int

const (
	_ AssetType = iota // 0 reserved to signal uninitialized
	AssetTypeChart // 1
	AssetTypeInsight // 2
	AssetTypeAudience // 3
)
```

and then either create some map(s) to have easy access to literals for marshalling or (even worse) to implement helper methods to switch between the literals. It looks to me that despite the obvious disadvantages of e.g. no compiler enforcement of the AssetType legal values, I would stick to the string...

internal/config/config.go
dzacharakis: Correct — but it’s due to one of Go’s key features: variables in Go can’t be uninitialized thanks to its _zero-value mechanism_.

internal/handler/handler.go
dzacharakis: yes, I was looking for a `linter`, tool that helps ensure the code adheres to the defined style guidelines and potential bugs.

internal/domain/asset.go
dzacharakis: In this case, you’ve created a new named type that shares `string` as its underlying type. Usually, the purpose of such a named type is to differentiate between distinct or incompatible usages of the `string` (here) type, ensuring they aren’t accidentally interchanged.

Nice alternative (enum), especially since it’s a defined set, but what I was asking about was an _interface type_ for a more general use case—similar to what you mentioned in your other comment.

internal/server/server.go
dzacharakis: I agree with your approach, and it’s good that you’re familiar with `PATCH`. I would use it on an endpoint like `/assets/{assetId}`, permitting updates only to fields like `description` or other relevant asset attributes passed in the `PATCH` request body.

dzacharakis:
Thank you, Gianni, for responding to my comments. 
All good on my end. Good luck with the next steps, and happy Christmas holidays!

PR #54, opened 2025-11-26
GWI Interview Task

bsakel:
Hello reviewer,
this is my idea of how someone could implement the task. I should point out that I am a complete golang novice but I tried to combine traditional studying (tutorials, documentation, etc) and some AI with Claude.

I wanted to build mainly a graphql api because in my opinion it fits better with the needs of a ui. However I built a rest api for the majority of the crud operations but not for the main request of the task.  In my readme you can find more information on the libraries used and links to the rest of the documentation.

I focused more on the completeness, with tests and a docker compose file, and less on the performance optimization. That said, there are some optimization ideas in the readme and there is another branch in my fork that has the implementation for each one (linked in the readme) and there are tests that can be used to measure the performance. In any case, this implementation is by no means a complete production ready one, and even if we considered the business logic as complete, at the very least it lacks proper logging, auth and observability.


.env_example
```diff
@@ -0,0 +1,2 @@
+DB_URL=<REPLACE THIS WITH YOUR DATABASE URL>
+TEST_DB_URL=<REPLACE THIS WITH YOUR DATABASE URL>
```
pamanta, line 2: Is the extra var needed here? Imagine we need to deploy this to `dev`, `staging`, `test` and `production` environments. How would you configure the database connection for each?

api/audience_handlers.go
```diff
@@ -0,0 +1,81 @@
...
+		return
+	}
+	db.GormDB.Create(&audience)
+	model.ResponseJSON(c, http.StatusCreated, "Audience created successfully", audience)
+}
+
+func GetAudiences(c *gin.Context) {
+	if db.GormDB == nil {
+		log.Fatal("DB pointer is nil")
+	}
```
pamanta, line 29: Would this be OK for a production environment?

docs/API_DOCUMENTATION.md
```diff
@@ -0,0 +1,334 @@
...
+
+### Audiences
+
+| Method | Endpoint | Description |
+|--------|----------|-------------|
+| POST | `/audience` | Create a new audience |
+| GET | `/audiences` | Get all audiences |
+| GET | `/audience/:id` | Get audience by ID |
+| PUT | `/audience/:id` | Update audience by ID |
+| DELETE | `/audience/:id` | Delete audience by ID |
```
pamanta, line 13: Would you consider changing it to something more restful, what could be the benefits and drawbacks of it?

models/userstar.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+	return nil
+}
+
+type UserStar struct {
+	ID      uint      `json:"id" gorm:"primaryKey"`
+	UserID  uint      `json:"userid"`
+	Type    AssetType `json:"type"`
+	AssetID uint      `json:"assetid"`
+}
```
pamanta, line 68: I noticed that the `AssetID` can reference different tables based on type. Can you walk me through the rationale behind this design decision? Did you consider any other solutions?

pamanta (commented):
Hello @bsakel

Thank you very much for taking the time to complete the assignment :).
Your idea to add GraphQL was really interesting!

I've left some comments for you to have a look when you have the time

.env_example
bsakel: The .env file is there only for development, and the idea is that a dev might want to keep two dbs one clean for automated testing and one with data for manual testing. In the case of other environments, I expect that those would be supplied from environmental variables (there is a provision for that).

api/audience_handlers.go
bsakel: Not really. The idea here is that this is a singleton and without it having a value something went terribly wrong and we need to fail the entire process. A more robust implementation would account for transient errors and avoid fatal errors because they fail the entire process. That said, even with the fatal errors we could make it work by restarting the process at a higher level like k8s.

models/userstar.go
bsakel: It might be wrong but my assumption is that the assets are common between all customers, and each customer chooses which they care more about and star only those. So one very simple way to do that is have a table to store the star with the user, asset id and the asset type. There could be other ways to accomplish the same but it felt good enough for this case.

I did not consider other options because of time but I briefly though around what the implications would be if the assets were not personalized per user.

docs/API_DOCUMENTATION.md
bsakel: Sorry I am not sure what you mean. Even though I added a rest api for the adding, deleting, updating and getting of all the asset types, it is in general there for completeness because I cared more for the graphql.

docs/TESTING_GUIDE.md
```diff
@@ -0,0 +1,242 @@
...
+
+**Purpose:** Measure query performance under different data loads
+
+**Available Benchmarks:**
+
+| Benchmark | Description | Data Size |
+|-----------|-------------|-----------|
+| `BenchmarkUserStared_SmallDataset` | Typical user load | 5 items per user |
+| `BenchmarkUserStared_MediumDataset` | Average user load | 50 items per user |
+| `BenchmarkUserStared_LargeDataset` | Heavy user load | 200 items per user |
```
lkslts64, line 212: Looking at your performance benchmarks for BenchmarkUserStared_SmallDataset, BenchmarkUserStared_MediumDataset, and BenchmarkUserStared_LargeDataset, how do the results scale as the dataset size increases? Are the performance characteristics what you expected?

main.go
```diff
@@ -0,0 +1,73 @@
...
+	router.GET("/userstars", api.GetUserStars)
+	router.GET("/userstar/:id", api.GetUserStar)
+	router.PUT("/userstar/:id", api.UpdateUserStar)
+	router.DELETE("/userstar/:id", api.DeleteUserStar)
+
+	// GraphQL routes
+	router.POST("/graphql", graphqlHandler(resolver))
+	router.GET("/graphql", playgroundHandler())
+
+	router.Run(":8080")
```
lkslts64, line 72: Currently, your application calls `router.Run(":8080")` without any signal handling. What happens when the application receives an OS signal like SIGTERM or SIGINT (e.g., during a deployment or container restart)? How would you implement graceful shutdown to ensure in-flight requests complete and resources are cleaned up properly? And finally, why graceful server shutdown is important when an app is deployed on Kubernetes?

main.go
```diff
@@ -0,0 +1,73 @@
...
+
+func playgroundHandler() gin.HandlerFunc {
+	h := playground.Handler("GraphQL", "/graphql")
+
+	return func(c *gin.Context) {
+		h.ServeHTTP(c.Writer, c.Request)
+	}
+}
+
+func main() {
```
lkslts64, line 30: If you were to deploy this application to production, what additional features, improvements, or infrastructure would you prioritize? Please briefly outline the key areas you'd focus on (e.g., observability, security, reliability, scalability).

lkslts64 (commented):
Hey @bsakel , thank you for taking the time to complete this challenge and for the effort you've put into your submission. We really appreciate it!

You did a great job, especially when having in mind that you're new to Go. 

I really liked the comprehensive testing suite you added, your GraphQL implementation in addition to the REST one and finally, the rich documentation you provided 💯 

Below, I have left some comments. Please take a look when you find some time ☺️

main.go
bsakel: To be honest, I think this is a very naive implementation for a production system. For instance, I am not using any dependency injection so everything is pretty hardcoded and there is limited capability to change implementations for eg unit testing. Another thing is that I am writing to the db inside the main code of the handler which I think needs some work, at least in order to handle transient errors. 

Besides that I would for sure start with logging as it is the simplest way to see what the app is doing or failing to do. Then tracing is a good way to analyze issues and solve them.

In terms of security I have done exactly zero! So a good start would be to add token authorization and some checks that the user actually has rights to do what they ask to do.

As far as I can tell, the star part of the system, is heavy on reading so adding a cache layer or even better precompiling projections of the most used assets that can be cached in cloudflare or another similar cdn, would help reduce the db load and help with scalability.

main.go
bsakel: Again being honest, I didn't think of this or knew it but I investigated it a bit. 

So as I understand it in k8s before restart it sends a signal for the app to gracefully stop. I believe this would be really important to handle especially if we have used concurrency to write or update. 

In my implementation I am not sure it changes much to be honest as the write or update should be really fast and batching is not implemented (in this case we would need flush to the db any buffers we have), however adding some type of handling of OS signals would be a valid reason for refactoring the way the handlers are build.

docs/TESTING_GUIDE.md
bsakel: The idea behind the performance tests was not so much to test the app with different datasets but mostly to be able to get a baseline before optimizations and after optimizations. 

Since I basically lost the results from when I did the exercise I rerun the test and I got 

Benchmark | Iterations | Time/op | Memory/op | Allocs/op
-- | -- | -- | -- | --
SmallDataset (5 items) | 135 | 9.06 ms | 2.72 MB | 60,777
MediumDataset (50 items) | 118 | 10.14 ms | 3.28 MB | 73,467
LargeDataset (200 items) | 81 | 14.13 ms | 5.00 MB | 111,603
MinimalFields | 91 | 13.72 ms | 4.27 MB | 98,681
AllFields | 74 | 14.23 ms | 4.99 MB | 106,900
Parallel | 274 | 3.73 ms | 2.99 MB | 64,740

Which to me has two interesting things 
1) Small to Medium seems to be bound by some overhead that we will pay regardless of the size of the test, so the time for 5 items is only 1 ms less on average than the time for 50. Medium to Large seems a bit more linear.
2) MinimalFileds vs AllFields seems really insignificant for the server (ok I know this example of a server is barely a usable one but still).

I am now sure if you have seen it but there is another branch in my repo with some optimization implementations, and I did execute the same tests there and got 

Benchmark | Before | After | Improvement | Memory Before | Memory After | Memory Saved
-- | -- | -- | -- | -- | -- | --
SmallDataset | 9.06 ms | 5.29 ms | 41.6% faster | 2.72 MB | 120 KB | 95.6% less
MediumDataset | 10.14 ms | 6.40 ms | 36.9% faster | 3.28 MB | 628 KB | 80.9% less
LargeDataset | 14.13 ms | 10.74 ms | 24.0% faster | 5.00 MB | 2.59 MB | 48.2% less
MinimalFields | 13.72 ms | 10.48 ms | 23.6% faster | 4.27 MB | 1.91 MB | 55.3% less
AllFields | 14.23 ms | 11.16 ms | 21.6% faster | 4.99 MB | 2.44 MB | 51.1% less
Parallel | 3.73 ms | 2.51 ms | 32.7% faster | 2.99 MB | 1.25 MB | 58.2% less

My observation here is that besides the expected to improvement in speed, the main 2 notes from before remain (Small is mainly governed by overhead and the gains by the limiting the fields are less than expected).

Two notes on the tests 
1) they have a bug in the cleanup step and the user stars are not cleared properly (the delete statement was not updated for user_favourites).
2) Testing with docker is very hit or miss, I did a couple more executions and the results were totally the opposite with significant slowdowns everywhere mainly due to the concurrency optimization overpowering the db.

PR #61, opened 2025-12-16
CBO Favorites Iplementation

cboursinos:
# GWI Platform Go Challenge - Favorites API

A high-performance, production-ready REST API for managing user favorites (assets) built with Go. This solution demonstrates Go programming concepts including concurrency, security, performance optimization, and clean architecture.

## Author CBO

Built as a demonstration of Go engineering skills, focusing on:
- Clean architecture and code organization
- Small extra Business Logic (Favorite Lists, Asset Managment)
- Performance optimization
- Security best practices
- Production-ready features
- Comprehensive testing


internal/api/middleware.go
```diff
@@ -0,0 +1,157 @@
...
+				respondWithError(w, http.StatusInternalServerError, "internal server error")
+			}
+		}()
+		next.ServeHTTP(w, r)
+	})
+}
+
+// CORSMiddleware handles CORS
+func CORSMiddleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
pamanta, line 128: Can you please elaborate a bit on what is the purpose of this middleware and why do we need to add it as part of this assignment?

internal/storage/cached_storage.go
```diff
@@ -0,0 +1,316 @@
...
+	createdAsset, err := s.storage.CreateAsset(userID, asset)
+	if err != nil {
+		return nil, err
+	}
+
+	// Invalidate asset cache
+	assetCacheKey := cacheKeyAsset(asset.GetID())
+	if err := s.cache.Delete(assetCacheKey); err != nil {
+		fmt.Printf("Warning: failed to invalidate asset cache for key %s: %v\n", assetCacheKey, err)
+	}
```
pamanta: Why is cache invalidation needed upon asset creation?

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN go mod download
+
+# Copy source code
+COPY . .
+
+# Build the application
+RUN CGO_ENABLED=0 GOOS=linux go build -a -installsuffix cgo -o server ./cmd/server
+
+# Final stage
+FROM alpine:latest
```
pamanta: Can you identify any drawbacks of using `latest` for image build?

internal/storage/storage.go
```diff
@@ -0,0 +1,40 @@
+package storage
+
+import (
+	"github.com/gwi/platform-go-challenge/internal/models"
+)
+
+// Storage defines the interface for storage operations
+type Storage interface {
```
pamanta, line 10: Looking at the Storage interface, I'm curious about your thoughts on request lifecycle management. In a production environment, how would you handle scenarios where a client disconnects or a request needs to be cancelled? What changes might you consider to the interface to support this?

go.mod
```diff
@@ -0,0 +1,31 @@
+module github.com/gwi/platform-go-challenge
+
+go 1.18
```
pamanta: Can you elaborate a bit on the decision of using an older version of Go?

internal/service/favorites.go
```diff
@@ -0,0 +1,451 @@
...
+
+	default:
+		return fmt.Errorf("unknown asset type: %s", asset.GetType())
+	}
+
+	return nil
+}
+
+// CreateAssetFromJSON creates an asset from JSON data
+func (s *FavoritesService) CreateAssetFromJSON(data map[string]interface{}) (models.Asset, error) {
```
pamanta, line 385: I see that you're manually extracting fields from the map. Have you considered using `json.Unmarshal` directly? What trade-offs do you see between the current approach and using the standard library's JSON unmarshaling?

internal/api/handlers.go
```diff
@@ -0,0 +1,733 @@
...
+	if req.Name == "" {
+		respondWithError(w, http.StatusBadRequest, "list name is required")
+		return
+	}
+
+	list, err := h.service.CreateList(userID, req.Name)
+	if err != nil {
+		respondWithError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
```
pamanta: This will return a `500` error, for all the errors that are originating from the service layer. Is this a good approach? If not how would you improve it in a production environment?

pamanta (commented):
Hi @cboursinos 👋.

Thank you very much for taking the time to complete the assignment!
Your submission had a lot of interesting features, but there are some key components/files missing from it.

I've left some comments for you to have a look when you have the time!

internal/api/middleware.go
cboursinos: I added this as an extra layer of protection as you requested about security! An extra level not call the app without the right access. The JWT for users but also we need for our app only!

internal/storage/cached_storage.go
cboursinos: I haded in the start when I had a small different implementation and forgot to remove it when I changed it

Dockerfile
cboursinos: Of course you can brake the “world”. I left as it is for this implementation! You never go to the latest. Think about having a bug and you need to deploy and find out that your pipeline build fails because a latest release misses packages.

go.mod
cboursinos: Ohhh you are right. I used a small project that had it in the past about reading and studying go. And left it as it is! Of course we can use the latest for our environment like the latest 25

internal/api/handlers.go
cboursinos: Yes you are right and here is a good question. The best approach is to design an error handling system with typed errors and centralized handling. We need to handle different errors with different error codes.  500 error code for server-side errors. 4xx codes for client-side errors.

internal/storage/storage.go
cboursinos: Yeap. I haven't implement context propagation: methods don't accept context.Context, so they can't respond to Client disconnections, Request timeouts, Graceful shutdown, Cancellation signals. In order to solve it, we need to add context.Context as the first parameter to all Storage interface methods. 
Update implementations to use context-aware database methods:
QueryContext() instead of Query()
QueryRowContext() instead of QueryRow()
ExecContext() instead of Exec()
BeginTx(ctx, ...) for transactions
Propagate context from HTTP handlers → services → storage.
Add request timeout middleware to enforce per-request timeouts.
Handle context errors appropriately (cancelled vs. deadline exceeded).

cboursinos:
@pamanta thank you for your review. I appreciate it. Sorry for my late replies, I am abroad with my family for my annual Christmas holidays. For me this were a test how fast I can embrace a new technology. By doing this project I wanted to share more my experience in different aspects, such as, product (new functionality based on the request), security, caching layers, implementation, testing, KPIs. Thank you for your time. Merry Christmas and a Happy New Year.

internal/service/favorites.go
cboursinos: In this solution we do not do double marshaling: because it works directly with the map, its flexible: handles missing optional fields and it's explicit: you see exactly what's extracted. On the other hand we can use a two-step approach: First unmarshal to a lightweight wrapper to get the type field and then unmarshal to the specific struct based on type (this is the two times marshaling). we can use a standard library as intended, automatic type conversion: handles string→int, float64→int, etc. Better error messages: detailed parse errors from the JSON library. And for sure we would had less code.

cboursinos:
Hi @pamanta happy new year, I pushed an updated version to the test!

pamanta:
Hi @cboursinos thanks, happy new year!

PR #63, opened 2025-12-16
Simple implementation

apostoloss:
Includes simple rest api, swagger, a health endpoint and simple gh-action.


alesr:
Hi @apostoloss, thank you for submitting your solution for the assignment. We’re currently reviewing your pull request and will get back to you as soon as possible.

internal/api/handlers.go
```diff
@@ -0,0 +1,192 @@
...
+// RegisterHandlers sets up all API routes on the provided router.
+func (api *API) RegisterHandlers(r *mux.Router) {
+	// Browse available assets (catalog)
+	r.HandleFunc("/assets", api.listAssetsHandler).Methods("GET")
+	r.HandleFunc("/healthz", healthHandler).Methods("GET")
+	r.HandleFunc("/users/{id}/favorites", api.listFavoritesHandler).Methods("GET")
+	r.HandleFunc("/users/{id}/favorites", api.addFavoriteHandler).Methods("POST")
+	r.HandleFunc("/users/{id}/favorites/{assetID}", api.removeFavoriteHandler).Methods("DELETE")
+	r.HandleFunc("/users/{id}/favorites/{assetID}", api.editFavoriteHandler).Methods("PATCH")
+}
```
alesr, line 29: In your API design, you're using the asset ID in the URL path for favorite operations:

```
DELETE /users/{id}/favorites/{assetID}
PATCH /users/{id}/favorites/{assetID}
```

Can you walk me through your reasoning for this approach? 

What led you to identify favorites by their associated asset ID rather than giving each favorite its own unique identifier?

apostoloss: By giving each favorite an id different that the assetid (own and unique identifier) we allow (if not correctly implemented) to favorite the same asset multiple times (with different descriptions as well). So the easiest way to prevent this is to reuse the assetid. 
Assuming (I think it is in the original Readme) that a user can have any number of favorites, there is a possibility for a user with huge number of favorites to experience delays adding a new favorite if the duplicate prevention mechanism had to lookup for assetIDs in their current list of favorites.

alesr: Hi @apostoloss, thanks for answering the question.

The benefits you mentioned: no duplicates, and efficient lookups, happen naturally with proper database design (composite key on `user_id, asset_id`). With indexing, duplicate checking is just constraint validation, not expensive scanning.

REST principles set you up for scale. It forces us to think about the domain model upfront. 

For example, if we later we need to extract favorites into its own microservice, this design translates beautifully. The favorites service can still expose the same resource model, and client code doesn't change.

cmd/server/main.go
```diff
@@ -0,0 +1,62 @@
...
+// @title Favorites API
+// @version 0.1
+// @description API for managing user's favorite assets
+// @host localhost:8080
+// @BasePath /
+func main() {
+	// Get configuration from environment
+	port := getEnv("PORT", "8080")
+	catalogPath := getEnv("CATALOG_PATH", "sample_data/seed_assets.json")
+	instanceID := getEnv("INSTANCE_ID", "default")
```
alexandros-georgantas, line 26: Could you please describe what is the purpose of the `instanceID`? How are you using it in your implementation?

cmd/server/main.go
alexandros-georgantas, line 1: I don't see any graceful handling of server errors or shutdown. If you had more time how would you have handled that?

internal/models/favorite.go
```diff
@@ -0,0 +1,20 @@
+package models
+
+import "time"
+
+// Favorite represents a user's reference to a favorited asset.
+// It stores only the asset ID and user-specific metadata, not the full asset.
+type Favorite struct {
```
alexandros-georgantas, line 7: I don't see any correlation with a a user at this level. Could you elaborate your decision of ditching `user_id` from this model?

internal/store/store.go
alexandros-georgantas, line 1: Could you please discuss the reason why there are inconsistencies in the return statements of these methods? Would you consider returning more than a potential error in some cases e.g. the new favorite entry created for the case of `AddFavorite` ?

internal/api/handlers.go
alexandros-georgantas, line 1: If you had more time how you would have refactored this file in order to be more readable?
Would you consider having consistently predefined return structures for your handlers?

alexandros-georgantas (commented):
Hello @apostoloss 👋  and Merry Christmas 🎄
 
Thank you for your time implementing this assignment!
I've also left some comments on your PR and would like to hear your thoughts!

Thanks,
Alex

